PR #200, opened 2018-08-23
 Moved ActivityMapper to own project for reusability

RaphaelRo:


src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
```diff
@@ -175,7 +190,7 @@ private static ActivityCategory GetCategory(string processName, string windowNam
             }
             catch (Exception e)
             {
-                Console.WriteLine("> ERROR while Mapping: {0}", e.Message);
+                //Console.WriteLine("> ERROR while Mapping: {0}", e.Message);
```
Hirse: Why not log this?
Should this be treated in any way?

src/client/AM.PA.MonitoringTool/WindowsActivityTracker/Visualizations/DayFragmentationTimeline.cs
```diff
@@ -6,11 +6,13 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
Hirse: I only have minimal C# experience, but adding a `using`-statement without changing the code seems incorrect. Is this import really used?

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
```diff
@@ -71,15 +70,19 @@ private static ActivityCategory GetCategory(string processName, string windowNam
...
             {
                 if (string.IsNullOrEmpty(processName) && string.IsNullOrEmpty(windowName))
+                {
                     return ActivityCategory.Unknown;
+                }
 
-                if (windowName != null) windowName = windowName.ToLower(CultureInfo.InvariantCulture);
-                if (processName != null) processName = processName.ToLower(CultureInfo.InvariantCulture);
+                windowName = windowName?.ToLower();
+                processName = processName?.ToLower();
```
Hirse: Why is the parameter `CultureInfo.InvariantCulture` no longer required?

src/client/AM.PA.MonitoringTool/ActivityMapper/ActivityCategoryExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Linq;
```
Hirse: Why does this file not have the license notice as the one above?

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
RaphaelRo: You don't have a console window open in a wpf application. This usually goes nowhere (except in the debugger I think). Viable alternatives are Trace.WriteLine, Debug.WriteLine or some other logging method. This is why I only commented it out.

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
RaphaelRo: The project I created is .NET Standard (as opposed to .NET Framework) and the version I used imho does not even know about culture or at least does not provide the ToLower functions using Cultures. This is why I had to change the code here.
Cultures most certainly only matter in interpretable data like DateTime, float, double or decimal. For windowName and ProcessName it wouldn't. The method without overload would default to InvariantCulture anyway.

src/client/AM.PA.MonitoringTool/WindowsActivityTracker/Visualizations/DayFragmentationTimeline.cs
RaphaelRo: You're absolutely right. This stems from a code change I didn't push in the end. My bad.

src/client/AM.PA.MonitoringTool/ActivityMapper/ActivityCategoryExtensions.cs
RaphaelRo: I would prefer to stay semi-anonymously. I added a license notice without a name. You're free to use yours.

Hirse: This might be a question for @casaout: What is your policy on license headers? Should every file have one?

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
Hirse: @casaout How would you handle these errors? Do you have a log file?

src/client/AM.PA.MonitoringTool/WindowsActivityTracker/Models/WindowsActivity.cs
```diff
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using ActivityMapper;
```
Hirse: Is this import used?

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
casaout: Depending on the error, we log issues into the log-table in the database. Database.GetInstance().LogError(…) -- Like that you can also log warnings and Infos.

For crazy things that make the tool/tracker unusable, we log them in the logfile Logger.WriteToLog(e); // here you can just add the exception directly

src/client/AM.PA.MonitoringTool/ActivityMapper/ProcessToActivityMapper.cs
casaout: Good to know, didn't know that. I've added them everywhere after the Microsoft Static Code Analysis tool suggested that 😁

src/client/AM.PA.MonitoringTool/WindowsActivityTracker/Models/WindowsActivity.cs
RaphaelRo: yes. it is used for the `ActivityCategory ` enum in `public ActivityCategory ActivityCategory { get; set; }`

casaout:
Thanks Raphael for suggesting to extract the ProgramToActivity-Mapper to a separate project. I just talked to @Hirse and he suggested that (if there is enough demand to use the mapper separately), he would create a new repository, instead of having it as a separate project here. Using Nuget, we could easily make it available to our projects (and others too).

Would you be interested in implementing this? (I would set-up the repo and handle the packaging using Nuget)

RaphaelRo:
Sounds good to me 👍

casaout:
Please move the project to https://github.com/sealuzh/ApplicationToActivityMapper.

Once it's ready and we've tested it, we can publish it to Nuget and integrate it into PersonalAnalytics.

PR #232, opened 2018-10-15
Execute a batch of queries in a transaction

Hirse:
This is based on the db-refatoring branch and includes those commits.
Please only look at the diff for the last commit (https://github.com/sealuzh/PersonalAnalytics/pull/232/commits/a2ae52475df93c6bcaf994ecf8d95459b2026881).
(It will look better once https://github.com/sealuzh/PersonalAnalytics/pull/230 is merged.)

### DIFF
--- a/src/client/AM.PA.MonitoringTool/Shared/Data/DatabaseImplementation.cs
+++ b/src/client/AM.PA.MonitoringTool/Shared/Data/DatabaseImplementation.cs
@@ -191,28 +191,20 @@ namespace Shared.Data
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Hirse:
Please don't merge yet.

I'll do a rebase and possibly add some more changes.
I'm still not sure whether the current function signature is the most useful. For batch inserts it is not necessary to repeat the query string, so I am think about `executeBatch(string, IEnumerable<object []>)`.

Hirse:
Rebased and updated to use only one query with a list of parameter values.

Issue #236, opened 2018-10-30
Reload TaskTray Icon when DisplaySettings Change


casaout:
On Windows, when the resolution/display changes, the task-tray icon sometimes appears blurry (e.g. when connecting and later disconnecting a separate monitor).

We could catch the [DisplaySettingsChanged](https://stackoverflow.com/questions/442337/detect-change-of-resolution-c-sharp-winforms)-Event and then re-load the icon.

PR #243, opened 2018-11-17
Reload icon on resolution change

Hirse:
See https://github.com/sealuzh/PersonalAnalytics/issues/236


casaout:
Awesome, thanks @Hirse

PR #245, opened 2018-12-04
Remove NuGet packages, closes #233

### LINKED ISSUE #233: Remove Nuget-Packages from being tracked
As suggested by @RaphaelRo in https://github.com/sealuzh/PersonalAnalytics/pull/230:
> Concerning the packages: Maybe you want to consider (in another task) removing the nuget obtainable packages from the repo. Although that will need some time to sort it out and test, it should unclutter the repo and slim it down my a substential margin.

Nuget-restore works a lot better/more reliable nowadays. HEnce, I think this is a great idea!

Hirse:
Updates the `.gitignore` with more general rules and removes all NuGet packages.
Restore of packages works fine on my machine, but maybe @casaout can have a look to make sure the server-side build is still working too.


casaout:
awesome, it works on my machine too, I will test it on Jenkins (after merging the PR).

casaout:
wow, had some serious issues with Nuget restore. Now it works again. Useful posts to solve the issue:
- https://stackoverflow.com/questions/48896486/jenkins-not-restoring-nuget-packages-with-new-msbuild-restore-target/48896487#48896487

PR #253, opened 2019-05-17
Lock batched inserts for concurrency

Hirse:
When writing batch-inserts from events I encountered interlocking transactions causing write problems.
This PR adds a lock so transactions are only started one at a time.


casaout:
thank you!

PR #259, opened 2019-11-19
Mac sqlite schema alignment

royru:
Hi all, 

This PR should align the db schema of the Mac version to the Windows version. Specifically, I removed all code related to Data Core, adapted the SQL queries to load data for the retrospective and re-implemented the tracker save functions to use sql INSERT statements. 

- The code to persist tracker data is still in `DataObjectController`. I did not want to make this PR any larger than it already is and will move save/persisting into the trackers themselves in a future PR.
- In AppDelegate I removed boilerplate code.
- All Core Data Classes have been removed (located in `/models`).

 **EmotionTracker**
- The table “EmotionalState” got renamed to emotional_state in order to conform with naming conventions. 
- the rest of the db schema is left unchanged.

**WindowActivityTracker**
- ActivitySQLController.getValidTimesQuery() was never used. I deleted it for the time being.
- I removed code related to "ZWEBSITE”.

**UserInputTracker**
- only V2 is implemented (i.e. `user_input` with minute-aggregates). Detailed tracking with tables `user_input_mouse_click` or `user_input_mouse_movement` is not yet supported. 
- I have yet to figure out how to retain events from special mouse buttons (i.e. middle button, left/right thumb button). 

**UserEfficiencyTracker**
- Currently, only the interval pop-up is implemented. Daily pop-ups are not yet supported.
- I downsized the pop-up to better match the Windows version (with the scale from 1-7). However, the windows version still has some neat features ("postpone" or "I did not work") which are missing.


csatterfield:
Sorry for the delay @royru I've had a busy last week. I'll try to get to this in the next few days. Thanks for your work!

src/client/AM.PA.MacOs/PersonalAnalytics/Shared/DataObjectController.swift
```diff
@@ -74,236 +83,124 @@ class DatabaseController{
 **/
 class DataObjectController: NSObject{
```
csatterfield: I think this whole class is kind of bad practice. Probably each tracker should have its own data controller class which is responsible for saving its data to the database. This way when we add more trackers we dont have to keep adding functions in this class.

Also let's move DatabaseController to it's own file.

src/client/AM.PA.MacOs/PersonalAnalytics/Trackers/WindowsActivityTracker/ActivitySQLController.swift
```diff
@@ -118,26 +98,28 @@ class ActivitySQLController: SQLController{
...
             
             if(row != nil){
-                let name: String = (row?["ZNAME"])!
-                let from:TimeInterval = (row?["ZSTARTTIME"])!
-                let to:TimeInterval = (row?["ZENDTIME"])!
-                let difference:TimeInterval = (row?["DIFF"])!
+                let name: String = row!["process"]
+                let from:TimeInterval = DateFormatConverter.dateStrToInterval1970(str: row!["tsStart"])
+                let to:TimeInterval = DateFormatConverter.dateStrToInterval1970(str: row!["tsEnd"])
+                let difference:TimeInterval = row!["diff"]
```
csatterfield: We should probably add constants for these field names (tsStart, tsEnd, etc..) somewhere as well as for the table names.

src/client/AM.PA.MacOs/PersonalAnalytics/Shared/SQLController.swift
```diff
@@ -64,18 +65,19 @@ class SQLController{
     
     func fetchAggregatedInputSince(time: Double) -> [AggregatedInputEntry] {
```
csatterfield: This function should be renamed to fetchUserInputSince, same with the AggregatedInputEntry class and anywhere else there remains a reference to aggregatedInput

src/client/AM.PA.MacOs/PersonalAnalytics/Shared/DataObjectController.swift
casaout: I agree. In the Windows version, we have a Queries class for each Tracker which stores all the queries that access the database.

royru: Totally agree. I already prepared this in the [mac-tracker-refactoring](https://github.com/sealuzh/PersonalAnalytics/tree/mac-tracker-refactoring) branch. I created a separate Queries class for each tracker and moved the methods in there.

src/client/AM.PA.MacOs/PersonalAnalytics/Shared/DataObjectController.swift
```diff
@@ -74,236 +83,124 @@ class DatabaseController{
...
-        lockQueue.sync {
-            let entity = NSEntityDescription.entity(forEntityName: "FocusState", in: managedContext!)
-            if let entity = entity{
-                let focus = NSManagedObject(entity: entity, insertInto: managedContext)
-                focus.setValue(Date().timeIntervalSince1970, forKey: "time")
-                focus.setValue(userInputLevel, forKey: "userinputlevel")
-                focus.setValue(focusState, forKey: "focusstate")
-                focus.setValue(smoothedFocusState, forKey: "smoothedfocusstate")
-            }
+    func saveUserInput(aggregatedInput input:UserInputTracker){
```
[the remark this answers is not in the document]
royru: "aggregatedInput" is the argument label when the function is called. I will remove it in the next branch.

src/client/AM.PA.MacOs/PersonalAnalytics/Trackers/WindowsActivityTracker/ActivitySQLController.swift
royru: Good point! Noted.

src/client/AM.PA.MacOs/PersonalAnalytics/Shared/SQLController.swift
royru: this function was used to fetch data for the csv export only. Since we discussed to remove the "export csv" functionality it is no longer needed. I already deleted it in the refactoring branch.

royru:
Thanks a lot for the feedback! It will be implemented in the next PR (branch is almost ready - I will submit the next PR once this one is merged).

csatterfield:
Sounds good! Merged.

PR #264, opened 2020-07-01
add python distribution to support PythonKit

csatterfield:
Added a portable conda environment we can use so that we can use PythonKit without worrying about whether the end user has an available Python distribution or not. It's not an elegant solution, but it should work for now 🙂.

To use it:

`import PythonKit`

Before invoking the Python global for the first time, set environment variables as follows:

```
let pathToPythonHome = Bundle.main.path(forResource: "Python", ofType: "")!
let pathToPythonLib = Bundle.main.path(forResource: "Python/lib/libpython3.8", ofType: "dylib")!
        
setenv("PYTHON_LIBRARY", pathToPythonLib, 1)
setenv("PYTHONHOME", pathToPythonHome, 1)
```

Now you can use PythonKit as normal, e.g.:

```
        let sys = try Python.import("sys")
        print("Python Version: \(sys.version)")
```

Currently, the sklearn library and its dependencies are installed. To add new python libraries, navigate to the Python folder in terminal and run the following commands:

```
source bin/activate
conda-unpack
```

Now you can install packages using pip/conda as normal.


royru:
Hi Chris,

Does the commit comprise the whole Python language and the conda env? 10k files seem like a lot to push to GitHub. 
Would it be possible to automatically download and configure Python in a build step? Or maybe install it with CocoaPods? Or maybe [this](https://github.com/pvieito/PythonKit#swift-package-manager). Do you happen to know how TensorFlow for Swift ships PythonKit?

Let me know if you see an alternative to uploading all those files. 

Looking forward to using Python, though :)

Thanks,
Roy

csatterfield:
Hey Roy,

Totally agree 10k files is excessive. The majority of these files are Python language files and conda env files. PythonKit is already installed using Swift Package Manager, but it relies on the end user having a Python distribution installed on their machine and does not include its own (which is why I added in all these other files 🙂). Downloading and configuring it in a build step sounds like a good idea, but I probably wont have time to look into it.

Until this issue is resolved, its probably for the best we just work off this branch and don't merge it into the main mac branch.

Have you had a chance to try it out? Let me know if it works on your machine.

royru:
Hi Chris, 

Sounds good. Maybe I find time to look into that part. For now, we can keep the separate branch 👍

I am currently struggling to get the PythonKit module --> `No such module "PythonKit"`. I am not yet familiar with SPM. Do I have to run a command similar to `pod install`? Maybe you can point me in the right direction, otherwise I will do a bit more google research ;)

csatterfield:
I tried cloning a new repo to reproduce your issue, but I haven't been able to. You should just be able to build normally. Do you see the dependencies listed in the project navigator? 

image

Also, is the package listed in the project settings? (here):

image

Maybe try resetting the package cache - File > Swift Packages > Reset Packages Cache

royru:
Strange, now it works. Maybe I was impatient and the dependencies did not finish loading or I had to restart Xcode. I see the dependencies now in the navigator and it works great so far. Awesome work! Thanks for looking into it!

Have a nice weekend!

casaout:
@csatterfield do you want to merge the PR? Thanks for your work on this!

csatterfield:
@casaout We're going to work off the python branch for now until we've found a better solution, so we won't merge into the mac branch. I'll close this PR

PR #329, opened 2024-10-21
316 make the encryption optional

SRichner:
@casaout looking at this now, the emphasis on `(...single) *file was created* (based...)` seems a bit weird. What do you think?

image


casaout:
> @casaout looking at this now, the emphasis on `(...single) *file was created* (based...)` seems a bit weird. What do you think?
> 
> image

I agree - should we call it `an export file was created` or `a data-file` (related to how we refer to it elsewhere)?

src/electron/src/views/DataExportView.vue
```diff
@@ -219,8 +220,9 @@ function revealItemInFolder(event: Event) {
               <p>
                 Please click "Next" once you are ready to
                 <b class="dark:text-white">first review, and later share your data</b>. The export
-                that will be created with your permission in the next step will be encrypted and
-                password-protected.
+                that will be created with your permission in the next step will be
```
casaout, line 223: how will this read if dataExportEncrypted==false?

Isn't it:
> The export that will be created with your permission in the next step will be.

Should  we hide the entire sentence?

SRichner: good catch, updated so it hides the complete sentence

SRichner:
> > @casaout looking at this now, the emphasis on `(...single) *file was created* (based...)` seems a bit weird. What do you think?
> > image
> 
> I agree - should we call it `an export file was created` or `a data-file` (related to how we refer to it elsewhere)?

How about `Your data was exported and we created a file based on your preferences on the previous page. To share this file with the researchers, please take the following steps:`?

casaout (approved):

PR #347, opened 2025-01-13
fix: updated experience sampler to use 24-hour format

naman108:
I've updated the ExperienceSamplingView.vue to force a 24 hour date-time format.


casaout:
Thank you for your improvement, and sorry for the delay with reviewing it. I've made two minor comments.

PR #377, opened 2025-02-04
[#376] adjust file logging to only log 'silly' on debug

casaout:


src/electron/electron/config/logger.config.ts
```diff
@@ -9,5 +9,5 @@ export const LOG_FILE_NAME = {
 // see: https://github.com/megahertz/electron-log#log-levels
 export const LOG_LEVEL = {
   CONSOLE: 'silly' as LevelOption,
-  FILE: 'silly' as LevelOption
+  FILE: 'info' as LevelOption
```
royru, line 12: can you do the `is.dev` check here so you don't have the `LevelOption`'s  (silly and info) in the same place?

casaout: done, thanks for the suggestion. please re-review :)

Issue #352, opened 2025-01-24
Add Settings Page


casaout:
Create a settings/about window (inspired by FlowTeams) with the following pages:

- About (mostly as-is today)
  - please add a button to "Open app log files" and "Open collected data" (and remove them from the app context menu)
- Active Times (aka work/study rhythms): #332 
- Retrospection/AIRbar settings (if available)

Note that I suggest to reuse/repurpose/rename the About-page for the settings (assuming htis makes sense)

![Image](https://github.com/user-attachments/assets/9460e8db-97d7-460c-8cdb-f2e1629687de)

Issue #345, opened 2025-01-07
Uploader: Optimize Upload-Flow


casaout:
In a study, we had the problem that users followed the export wizard, but didn't read the text on the last page that explained how to actually upload/share the data. Thus, some data was lost.

Until we have an auto-upload to the Data Donation Lab (#324), we want to add a warning when we think that the upload did not yet complete:

When, on the last page of the wizard, the user tries to close the window ("x" or with "Close"-button), show a warning if the user did NOT yet click the two "Click here" (in steps 1 and 2).

The warning could be a pop-up or a colorful box within the page: 
`Warning: Note that the export was NOT yet uploaded and shared with the researchers. Please follow the three steps outlined on this page.`

Tbd, I wonder if, after showing the message, we should still allow to close the window?

naman108:
One alternative I was considering when first working with the software was how it could support a one click data upload instead of the final page with the three or so steps. By creating a minimal backend service that could be deployed on free tier cloud infrastructure which exposes a single endpoint for data upload so that participants (after removing any data they don't want visible) can have a one click data upload which is automatically aggregated.

casaout:
> One alternative I was considering when first working with the software was how it could support a one click data upload instead of the final page with the three or so steps. By creating a minimal backend service that could be deployed on free tier cloud infrastructure which exposes a single endpoint for data upload so that participants (after removing any data they don't want visible) can have a one click data upload which is automatically aggregated.

Thank you for your input, @naman108! That's definitely an idea we will consider more closely in the future. One challenge that we are facing in our particular deployments at University of Zurich are that we couldn't deploy an uploader to a free tier cloud storage service due to privacy reasons. Thus, for now, a first step to create an auto-upload feature (without the manual last step) will be to integrate with the universities' [Data Donation Lab ](https://datadonation.uzh.ch/en/) (#324). Non-UZH researchers can also use it, and we hope to have it ready in Q1'2025.

Issue #342, opened 2024-11-06
Visualize number of Days Participated in Study


casaout:
For studies, it might be valuable to visualize to participants how many days they've participated so far, given some kind of metric.

This idea is inspired by AwarenessBar's visualization of study-related information
![image](https://github.com/user-attachments/assets/9484a0f7-6e15-4535-a42e-f922d5ca4a4c)

Tasks:
- [x] create a new study config setting `displayDaysParticipated` (default: true)
  - [x] add a comment to point to the method and docu that calculates the measure, so it can easily be adapted (+ update in the documentation for researchers)
- [x] create a measure daysParticipated that is updated on a daily (?) basis
  - [x] as a simple default measure, we suggest to count as a study day if (a) at least one experience sampling response was given (if enabled) and (b) there are at least 60 user input entries (indicating 60+ minutes of activity)
- [x] visualize the number of days in the context menu: `Days participated: {number}`
- [x] also, while we're at it, display the subject id in the context menu: `Subject Id: {id}` and a add a `Copy Subject Id` button

Issue #332, opened 2024-10-25
Settings: Allow user to Configure Experience Sampling (with active/work hours)


casaout:
- [ ] generally, port the feature over from FT/AIRbar
- [ ] researcher can enable/disable the feature
- [ ] `tbd` user can enable/disable the feature (or only enable/disable experience sampling alltogether?)
- [ ] user can define "active hours"
- [ ] `tbd` user-setting where the user can decide whether to receive experience sampling pop-ups outside active hours
- [ ] `tbd` user-setting or researcher-setting where it's defined whether monitors run outside active hours

![image](https://github.com/user-attachments/assets/7ff6e677-40f1-447f-9f26-dca4442fbb03)

casaout:
@royru Discusssion notes:
- for now,we add a study-config setting where the researcher can decide whether the feature is enabled by default or not (default: true)
- update the documentation to explain the property
- later, we'll add it to the onboarding #381

PR #379, opened 2025-02-04
Settings / Work Hours / Data Export Warning / Days Participated

royru:
- Settings Page including "About" and "Active Times" (excluding the retrospection settings) [#352](https://github.com/HASEL-UZH/PersonalAnalytics/issues/352).
- Users can define active times (work hours) [#332](https://github.com/HASEL-UZH/PersonalAnalytics/issues/332). When enabled, experience sampling only works within work hours.
- During data export, closing the window will trigger a warning message unless the user opened the two links as part of the step-by-step guide on the last transition page [#345](https://github.com/HASEL-UZH/PersonalAnalytics/issues/345).
- The number of days that participated is shown in the tray menu [#342](https://github.com/HASEL-UZH/PersonalAnalytics/issues/342).
- Furthermore, this PR also includes the basis for the goal-setting/self-reflection component, which is currently disabled and its business logic in a private repo.


src/electron/electron/main/services/WindowService.ts
```diff
@@ -307,39 +345,65 @@ export class WindowService {
...
+        label: 'Copy Subject Id',
+        click: () => clipboard.writeText(settings.subjectId)  
+      },
+      {
+        label: `Subject ID: ${settings.subjectId}`,
+        enabled: false, 
+      }
     ];
+
+    if (studyConfig.displayDaysParticipated) {
```
casaout, line 372: i think we could simplify this, instead of the condition and then pushing the item to the list, we could add:

`visible: studyConfig.displayDaysParticipated`

(similarly below for the "open task planning"

src/electron/electron/main/services/WindowService.ts
```diff
@@ -307,39 +345,65 @@ export class WindowService {
...
       { label: `Version ${app.getVersion()}`, enabled: false },
       {
         label: 'Check for updates',
         enabled: false,
         click: () => this.appUpdaterService.checkForUpdates({ silent: false })
       },
-      { type: 'separator' }
+      { type: 'separator' },
+      { 
+        label: 'Copy Subject Id',
```
casaout, line 363: super minor, but i suggest to move the "copy"-feature below the "listing the subject ID" item

src/electron/electron/ipc/IpcHandler.ts
```diff
@@ -56,14 +71,19 @@ export class IpcHandler {
...
+      openUploadUrl: this.openUploadUrl,
       startAllTrackers: this.startAllTrackers,
       triggerPermissionCheckAccessibility: this.triggerPermissionCheckAccessibility,
       triggerPermissionCheckScreenRecording: this.triggerPermissionCheckScreenRecording
     };
-    LOG.debug('IpcHandler constructor called');
-  }
 
-  public init(): void {
+    if (studyConfig.trackers.taskTracker.enabled) {
```
casaout, line 80: for versions of PA where the taskTracker is unavailable, and thus, the config doesn't contain a related entry, will this just fail "silently"?

Also, I wonder if we should merge this to the official PA-repo, or keep in the fork?

casaout (changes requested):

casaout:
hi @royru 

Thanks so much for this amazing work!

I've made a couple of comments throughout the changes, and have a few more comments here:

1) I've noticed that with this version, all windows take extremely long to load initially. E.g. for the new settings window, it's 10s until the empty window is opened, and 20+ additional seconds until the content is loaded. the experience sampling and export window also take a while, but they're a bit faster. Checking the build on the dev-branch, the initial window creation is not instant, but much faster (<1s).

2) I didn't yet give feedback on the settings-UI, as I am assuming that we're first testing it. (there are a few minor suggestions that I'll have)

3) I just noticed that we don't add to the experience sampling table, if it was a triggered or manual experience sample (#380)

src/electron/electron/preload/index.ts
```diff
@@ -1,8 +1,38 @@
+// OTODO: how to deal with this??? - 2024-11-25
```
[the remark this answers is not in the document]
royru: I'll have to look into this piece of code again, still on my todo as it is related to the self-reflection component.

src/electron/electron/main/services/trackers/DaysParticipatedTracker.ts
```diff
@@ -0,0 +1,102 @@
...
+    } catch (error) {
+      LOG.error(`Error starting days participated counting job: ${error}`)
+      throw error
+    }
+  }
+
+  private async updateDaysParticipated() {
+    const settings: Settings = await Settings.findOneBy({ onlyOneEntityShouldExist: 1 })
+    // as a simple default measure, we suggest to count as a study day if (a) at least one 
+    // experience sampling response was given (TODO: if enabled) and (b) there are at
```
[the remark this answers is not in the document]
royru: yes, I mistakenly thought it's a user setting (we still would have to implement in the future, hence the "todo"), but just realized after reading your comment that it's actually a researcher setting which we can account for already. Will fix this.

src/electron/electron/ipc/IpcHandler.ts
royru: ha, good point. I assumed that every config will have this (set to `enabled:false` without the self-reflection compoment). I will make the taskTracker setting optional. And eventually we'll remove it from here completely when we do the fork.

royru: As a side note, instead of having the enabled flag on all the other trackers, we could also consider their existence in the config optional. This would simplify the code and the config setup for researchers. It'd be more like a plug-and-play thingy...

royru:
Thanks :)

1. as discussed in person, we check if this is better in the build and/or will soon extract the changes to the vite config to a fork. Hopefully, this will fix the issues.

2. Feel free to write an issue for the UI.

3. Good catch

src/electron/electron/main/services/WindowService.ts
```diff
@@ -275,20 +278,55 @@ export class WindowService {
...
     this.dataExportWindow.show();
 
-    this.dataExportWindow.on('close', () => {
+    this.dataExportWindow.on('close', async (event) => {
+      console.log("close event")
+      const seemsToHaveCompletedExport = this.hasOpenedDataExportUrl && this.hasRevealedDataEportFolder;
+      if (!seemsToHaveCompletedExport) {   
+        const result = await dialog.showMessageBox({
+          type: 'warning',
+          buttons: ['Cancel', 'Close Anyway'],
```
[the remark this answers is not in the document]
casaout: The wording is great.

It works better, but still has a bug:
1. close (without having clicked the two "click here")
2. in the warning select "continue with data export"
3. from now on, clicking "close" doesn't wrok anymore

fails with:
> 16:53:37.407 (WindowService)                > User chose to close data export window, closing...
(node:14168) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of null (reading 'destroy')
    at BrowserWindow.<anonymous> (file:///C:/DATA/DEV/UZH/PersonalAnalytics/src/electron/dist-electron/main/index.js:4268:29)
    at BrowserWindow.emit (node:events:517:28)
(node:14168) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)

src/electron/electron/main/services/WindowService.ts
```diff
@@ -265,131 +268,170 @@ export class WindowService {
...
+        label: 'Open Experience Sampling',
+        click: () => this.createExperienceSamplingWindow(true)
       },
       {
-        label: 'Open Collected Data', // todo: move to settings
-        click: (): void => {
-          LOG.info(`Opening collected data at ${app.getPath('userData')}`);
-          shell.showItemInFolder(path.join(app.getPath('userData'), 'database.sqlite'));
+        label: 'Open Task Planning',
+        visible: studyConfig.trackers.taskTracker?.enabled,
```
casaout, line 384: just to be sure as it's currently showing up for me: is taskTracker enabled for this test?

(i don't see it in study.config.ts)

casaout:
> Thanks :)
> 
> 1. as discussed in person, we check if this is better in the build and/or will soon extract the changes to the vite config to a fork. Hopefully, this will fix the issues.
> 2. Feel free to write an issue for the UI.
> 3. Good catch. Should we prioritize this or is it a nice to have?
> 
> After a final test on Windows we should be good to go.

1. sounds great, thanks!
2. done: #382
3. i'd say it's a nice to have for now (hopefully someone else picks up on it)

src/electron/electron/main/services/WindowService.ts
royru: sorry, my bad. Did not realize that undefined would default to "true" in this case since it's an optional property. Fixed now.

src/electron/electron/main/services/WindowService.ts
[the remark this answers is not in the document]
casaout: works even better, but the following use-case doesn't work:
1. try to close > receive warning > say continue > click both links
2. click close which should close the window now (but doesn't):  (a) clickcing "Close" doesn't do anything and (b) clicking "x" throws the following error:

> 14:02:58.731 (WindowService)                > User chose to close data export window, closing...
(node:6712) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of null (reading 'destroy')
    at BrowserWindow.<anonymous> (file:///C:/DATA/DEV/UZH/PersonalAnalytics/src/electron/dist-electron/main/index.js:4268:29)
    at BrowserWindow.emit (node:events:517:28)
(Use `electron --trace-warnings ...` to show where the warning was created)
(node:6712) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)

Issue #439, opened 2025-07-28
Update Packages


casaout:
On a separate branch, please update all used libraries to keep the project up-to-date.

For all the packages/libraries where the upgrade creates breaking changes, list them here so we can discuss the update path.

casaout:
@grigor-dochev While I've recently started with some of the most critical package updates, there are several critical ones that just came up; see [open PRs from dependabot](https://github.com/HASEL-UZH/PersonalAnalytics/pulls) and  [high/critical reports](https://github.com/HASEL-UZH/PersonalAnalytics/security/dependabot?q=is%3Aopen+severity%3Acritical%2Chigh).

A simple "fix" is usually to update the minor package versions that fix these issues only in the package-lock file (not the packages.json). Thtat's actually what dependabot does. I'd like to fix these asap as well; so may I ask you to try to update these packages that way (before we properly update the packages later on). There might be a few that create larger refactors/changes; so for now lets just update the easy ones. Thanks!

casaout:
Remaining major updates:

<img width="1967" height="978" alt="Image" src="https://github.com/user-attachments/assets/df7e8d02-e963-4971-bae1-021f175e944b" />

PR #452, opened 2025-10-13
[#439] updated npm packages

casaout:
I've updated the npm packages to the recommended ones using `npm updates` and made the required changes to some specifc css/tailwind-classes. It does compile for me, it runs (also stores data), and the key views load and seem okay.

Can you please check on your side / mac? Thanks!


royru:
@casaout it seems to work fine on mac. However, I don't see any updated packages in `packages.json`. Shouldn't the tailwind version be increased?

PR #456, opened 2025-10-27
[#347] Display Local Time in User's Preferred 12/24h Format

grigor-dochev:
## Summary
Improve time display in the Experience Sampling popup by respecting the user's locale (12h vs 24h format).

## Changes
- Use `Intl.DateTimeFormat` instead of substring slicing
- Detect locale preference for 12h or 24h
- More robust and consistent time formatting

## Example
| Before | After |
|--------|-------|
| `1:26:` | `13:26` / `1:26 PM` |


casaout:
I am not sure if it works correctly yet, as for me it shows 05:34PM, but per my locale it should be 17:34:

image


(let me know if you need any further info to debug this feedback)

casaout (changes requested):

casaout (approved):

PR #509, opened 2026-04-17
[#492] Add Daily Survey Tracker for start/end-of-workday questionnaires

### LINKED ISSUE #492: Create End-of-day / Start-of-day Questionnaire
In some studies, we pair the hourly experience sampling with either morning or evening questionnaires, that usually span several short questions.

Tasks:
- [ ] create a new "Tracker" called "dailySurveyTracker" named "Daily Survey" (originally, I wanted to extend the experienceSamplingTracker, but I think that's confusing)
- [ ] update the study config file to allow researchers to define evening and/or morning questionnaires
  - [ ] create a new section `dailySurveyTracker`
  - [ ] `samplingType`: `morning`, `evening` (note that researchers could poentially define a morning *and* evening survey)
  - [ ] `delay`: integer in minutes that uses either the start/end of workday +- the delay (e.g. when set to -30 for an evening survey; it will open 30 minutes before the planned end of the workday)
  - [ ] it then can have a questions tag and its contents can be the same as for experienceSamplingTracker (i.e. question, answertype, scale, responseOptions)
  - [ ] `tbd` should we definean order or do we take the one given by the researcher?
- [ ] update the database schema
  - [ ] add new table for dailySurvey
  - [ ] columns: created_at, updated_at, deleted_at, prompted_at, samplingType, question, responseOptions, scale, response, skipped, answerType
  - [ ] i suggest to store each response as an individual row and just ensure that the timestamps align so we can easily fetch the matching responses from one day
- [ ] create a separate window that shows the questions
  - [ ] show a stack of all questions (in the order they are defined in the study config file)
  - [ ] add option to postpone them by 5 minutes, 15 minutes and 60 minutes or skip
  - [ ] save button will save the results
  - [ ] usual close button will just close it (log as "skipped", similar to Experience Sampling)
  - [ ] trigger it at the start/end time (+- delay) that the user defined (i.e. so this requires that the active times setting is enabled)
    - [ ] `tbd` how does it behave when the computer is no longer running at the time it's supposed to be triggered? does it re-open the next time the computer runs? (if yes, do we need to show the date for when a person is providing a survey response?)
- [ ] update the documentation
- [ ] update the data donation

Example of how it looked in a recent study ([FlowTeams study](https://www.andre-meyer.ch/DATA/Pubs/2025_FlowTeams_CSCW25_PrePrint.pdf)):

Image

grigor-dochev:
# Add Daily Survey Tracker for start/end-of-workday questionnaires (Closes #492)

## Description

Adds a new DailySurveyTracker that allows researchers to define structured morning and/or evening questionnaires triggered relative to participants' configured work hours. Unlike the ExperienceSamplingTracker which shows a single randomly selected question at periodic intervals, the DailySurveyTracker presents all configured questions at once in a multi-question survey window.

### Changes Made

- New `dailySurveyTracker` configuration block in `study.config.ts` with support for multiple surveys, each tied to a `samplingType` (`morning` or `evening`) with a configurable `delayInMinutes` offset from work start/end
- `requireAllAnswers` configuration property: when `true` (default), participants must answer every question before the Save button becomes active; when `false`, partial submissions are allowed
- All four question types supported: `LikertScale`, `TextResponse` (singleLine/multiLine), `SingleChoice`, and `MultiChoice`, rendered in researcher-defined order
- Per-minute scheduling job that computes trigger times from the participant's work schedule, with persistence across app restarts
- Missed trigger handling: if the computer was off at the scheduled time, the survey fires on next launch, with a late submission banner shown if more than 12 hours have passed
- Postpone (5, 15, or 60 minutes) and Skip functionality, with postponed times persisted in the Settings entity
- Dynamic tray menu items: one entry per configured survey with the correct label
- Each question response stored as an individual row in a new `daily_survey_responses` table, grouped by a shared `promptedAt` timestamp
- Data export integration with a preview table of recent responses
- Usage data events for opened, completed, skipped, and postponed actions
- Updated documentation in README.md and RESEARCH.md

## Implementation Details

1. Trigger times are computed by reading the participant's work schedule, finding the next working day, and applying the `delayInMinutes` offset. Positive values delay past the boundary (e.g. `5` = 5 minutes after work start), negative values trigger before it (e.g. `-30` = 30 minutes before work end).
2. Next invocation dates are stored as two nullable datetime columns on the Settings entity (`nextDailySurveyMorningInvocation`, `nextDailySurveyEveningInvocation`), ensuring they survive app restarts and are not recomputed unnecessarily.
3. The `scheduledDate` is passed from the tracker through WindowService to the Vue view via URL query params, so late submission detection works even if the survey is opened long after its original schedule.
4. Each survey response is stored per-question (not per-survey) to keep the schema flat and queryable, matching the pattern used by ExperienceSamplingTracker.

## Screenshots

| Morning survey | Evening survey |
|:-:|:-:|
| Morning survey | Evening survey |

| Late submission banner | Data export |
|:-:|:-:|
| Late submission banner | Data export |

Closes #492


src/electron/src/components/DataExportDailySurveyTracker.vue
```diff
@@ -0,0 +1,90 @@
...
+      >
+        <thead class="border-b">
+          <tr>
+            <th>Sampling Type</th>
+            <th>Question</th>
+            <th>Answer Type</th>
+            <th>Response</th>
+            <th>Scale</th>
+            <th>Response Options</th>
+            <th>Skipped</th>
```
casaout: i think that we don't need to show skipped here (also not for experience sampling); so we can filter them out from the examples and remove the column. They will be exported, of course, but to the user, the actual responses are interesting

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private startCheckJob(): void {
+    this.checkJob?.cancel();
+    this.checkJob = schedule.scheduleJob('* * * * *', async () => {
+      const settings: Settings = await Settings.findOneBy({ onlyOneEntityShouldExist: 1 });
+      const now = new Date();
+
+      for (const survey of this.surveys) {
+        const invocationField = survey.samplingType === 'morning'
```
casaout: here (and elsewehre below), would it be cleaner to check the samplingtype for "evening" too and create an exception otherwise?

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
```diff
@@ -0,0 +1,147 @@
...
+      const fireTime = new Date(candidate);
+      fireTime.setHours(hours, minutes, 0, 0);
+      fireTime.setMinutes(fireTime.getMinutes() + survey.delayInMinutes);
+
+      if (fireTime > now) {
+        return fireTime;
+      }
+    }
+
+    const fallback = new Date(now);
```
casaout, line 123: what does the fallback do? maybe add an inline-comment?

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
```diff
@@ -0,0 +1,147 @@
...
+
+const weekDays = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday'];
+
+export class DailySurveyTracker implements Tracker {
+  private checkJob: schedule.Job;
+  private readonly windowService: WindowService;
+  private readonly workScheduleService: WorkScheduleService;
+  private readonly surveys: DailySurveyConfig[];
+
+  public readonly name: string = 'DailySurveyTracker';
```
casaout: should be called "Daily Survey" (see #492 )

src/electron/shared/study.config.ts
```diff
@@ -69,6 +69,84 @@ const studyConfig: StudyConfiguration = {
       allowUserToDisable: true,
       allowUserToChangeInterval: true,
       userDefinedInterval_h: [0.5, 1, 2, 3, 4]
+    },
+    dailySurveyTracker: {
```
casaout, line 73: Comment for Andre: refine these questions for the public testing

documentation/RESEARCH.md
```diff
@@ -81,6 +81,15 @@ You can find more information on electron-builder here: https://www.electron.bui
...
+|---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|---------------|
+| `enabled`                       | Whether the DailySurveyTracker is enabled. If enabled, the DailySurveyTracker will prompt the participant with the configured morning and/or evening surveys relative to their work hours.                                                                                                                                                                                                |                 | `false`       |
```
casaout, line 87: do we need to document requireAllAnswers?

src/electron/electron/main/services/WindowService.ts
```diff
@@ -448,6 +524,12 @@ export class WindowService {
         click: () => this.createExperienceSamplingWindow(true),
         visible: showSelfReportMenu
       },
+      ...(studyConfig.trackers.dailySurveyTracker?.enabled
```
casaout, line 528: can you create a condition so they're only shown on debug? (visible: is.dev)

src/electron/electron/main/services/WindowService.ts
```diff
@@ -132,6 +133,81 @@ export class WindowService {
...
+
+    UsageDataService.createNewUsageDataEvent(UsageDataEventType.DailySurveyOpened)
+
+    const __filename = fileURLToPath(import.meta.url)
+    const __dirname = dirname(__filename)
+    const preload = join(__dirname, '../preload/index.mjs')
+
+    const { width, height } = screen.getPrimaryDisplay().workAreaSize
+    const windowWidth = 750
+    const windowHeight = 700
```
casaout: when there are more than 2-3 questions, the window adds a scrollbar (which is great), but I think the MaxHeight of the window could be slightly larger (and ideally it adapts a bit to the content until it's reached a max height?

Image

src/electron/src/views/DailySurveyView.vue
```diff
@@ -0,0 +1,539 @@
...
+            class="choice-button"
+            :class="{ 'choice-button-selected': responses[index]?.value === option }"
+            :disabled="isSubmitting"
+            @click="setResponse(index, 'SingleChoice', option)"
+          >
+            {{ option }}
+          </button>
+        </div>
+
+        <div v-if="q.answerType === 'MultiChoice'" class="choice-container">
```
casaout, line 205: similar to the experience sampling pop-up, can we show when multiple selections are possible (i think when only one is possible, there is no need to show a message)

src/electron/shared/study.config.ts
```diff
@@ -69,6 +69,84 @@ const studyConfig: StudyConfiguration = {
...
+      enabled: true,
+      requireAllAnswers: true,
+      surveys: [
+        {
+          samplingType: 'morning',
+          delayInMinutes: 5,
```
casaout, line 79: not a change request, but to understand the feature a bit better. is the cronjob checking every minute whether the questionnaire should be shown?

Also, when the current one is postponed (e.g. by 60 minutes); i assume it's rescheduled for right now + 60min; when/how is the next day's experience sampling scheduled? (is this independent from answering the previous one?)

src/electron/shared/study.config.ts
```diff
@@ -69,6 +69,84 @@ const studyConfig: StudyConfiguration = {
...
+            {
+              question: 'How motivated are you to start today?',
+              answerType: 'LikertScale',
+              scale: 7,
+              responseOptions: ['not at all motivated', 'moderately motivated', 'very motivated']
+            },
+            {
+              question: 'What is your main goal for today?',
+              answerType: 'TextResponse',
+              responseOptions: 'singleLine',
```
casaout, line 96: can we also show the word counts/max here if it's set?

src/electron/shared/StudyConfiguration.ts
```diff
@@ -60,10 +60,25 @@ export interface ChoiceQuestion extends ExperienceSamplingQuestionBase {
...
+
+export interface DailySurveyConfig {
+  samplingType: DailySurveySamplingType;
+  delayInMinutes: number;
+  questions: ExperienceSamplingQuestion[];
+}
+
+export interface DailySurveyTrackerConfiguration {
+  enabled: boolean;
+  requireAllAnswers?: boolean;
```
casaout, line 73: the default behavior is `true` (if I saw that correctly).  I think it should be "false", i.e. not requried; as usually responses are better if participants are not forced to give them

casaout (changes requested):

src/electron/electron/enums/UsageDataEventType.enum.ts
```diff
@@ -11,5 +11,9 @@ export enum UsageDataEventType {
   ExperienceSamplingManuallyOpened = 'EXPERIENCE_SAMPLING_MANUALLY_OPENED',
   ExperienceSamplingAutomaticallyOpened = 'EXPERIENCE_SAMPLING_AUTOMATICALLY_OPENED',
   ExperienceSamplingAnswered = 'EXPERIENCE_SAMPLING_ANSWERED',
-  ExperienceSamplingSkipped = 'EXPERIENCE_SAMPLING_SKIPPED'
+  ExperienceSamplingSkipped = 'EXPERIENCE_SAMPLING_SKIPPED',
+  DailySurveyOpened = 'DAILY_SURVEY_OPENED',
+  DailySurveyCompleted = 'DAILY_SURVEY_COMPLETED',
+  DailySurveySkipped = 'DAILY_SURVEY_SKIPPED',
+  DailySurveyPostponed = 'DAILY_SURVEY_POSTPONED'
```
casaout, line 18: when postponed, I think we should log in `additionalInformation` how many minutes it was postponed in json format)

src/electron/src/views/DailySurveyView.vue
grigor-dochev: I've added a message for the multiple selection questions.

Screenshot 2026-04-20 at 11 41 57

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
grigor-dochev: If no working day was found in the next 7 days, it schedules for tomorrow at 9am.

src/electron/shared/study.config.ts
grigor-dochev: Exactly, the cron job checks every minute. When you postpone, the invocation time is set to now + N minutes. The next day's survey is scheduled independently, so it's computed when the current one fires, which means that answering/skipping doesn't block the next day's scheduling. Does this logic make sense, or do you think we need to change it?

casaout: yes, this sounds perfect!

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
casaout: In which case could there be no working day in the next 7 days? Basically, if I do this?
image

Not saying it's wrong, but I'll have to understand it better:
What is the thinking of then showing it the next day at 9am?

This raises another question: what happens if no active times are set? is this even possible?
image

src/electron/electron/main/services/WindowService.ts
casaout: here, for example, the height is too much:
image

casaout (changes requested):

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
grigor-dochev: Yes, it's possible, if the participant toggles all 7 days off in their work schedule settings. In that case, `computeNextInvocation` will find no working day in the next 7 days and so it falls back to scheduling the survey for tomorrow at 9am (with a warning logged). This repeats daily, so the survey still fires on a default schedule rather than silently never appearing.

Also, the `WorkScheduleService` returns a default Mon-Fri schedule if no work days have been saved at all, so this fallback only triggers when the participant has explicitly disabled every day.

casaout: i like it, thanks for explaining!

I just realized that now the setting `enabledWorkHours` which so far could only be set in experiencesamplinggracker doesn't make sense. i think we need to elevate it one level (also to align with `showActiveTimesInOnboarding`)

What do you think?

src/electron/shared/StudyConfiguration.ts
casaout: btw, i just noticed that this is defined on a tracker level (i.e. same for morning and evening survey); 
i think it's better to be able to define this individually

src/electron/src/views/DailySurveyView.vue
```diff
@@ -0,0 +1,567 @@
...
+
+    <div class="postpone-bar">
+      <button class="postpone-button" :disabled="isSubmitting" @click="postpone(5)">Remind me again in 5 minutes</button>
+      <button class="postpone-button" :disabled="isSubmitting" @click="postpone(15)">Remind me again in 15 minutes</button>
+      <button class="postpone-button" :disabled="isSubmitting" @click="postpone(60)">Remind me again in 60 minutes</button>
+      <button class="postpone-button" :disabled="isSubmitting" @click="skipSurvey()">Skip</button>
+    </div>
+
+    <div v-if="isLateSubmission" class="late-notice">
+      This survey was originally scheduled for {{ scheduledDateString }}.
```
casaout, line 141: i wonder if there is a bug with the logic, as it shows me that the survey was scheduled for April 20th; whereas i just triggered it manually from the conext menu. (also my active work hours are fine)

Image

src/electron/electron/main/services/WindowService.ts
casaout: When you have fewer questions, does the window height shorten?

For me it's still very high:

image

casaout (changes requested):

src/electron/electron/main/services/DatabaseService.ts
```diff
@@ -9,6 +9,7 @@ import { UserInputEntity } from '../entities/UserInputEntity';
 import { Settings } from '../entities/Settings';
 import { UsageDataEntity } from '../entities/UsageDataEntity';
 import { WorkDayEntity } from '../entities/WorkDayEntity';
+import { DailySurveyResponseEntity } from '../entities/DailySurveyResponseEntity';
 
 const LOG = getMainLogger('DatabaseService');
```
casaout, line 14: I have a few suggestions to improve the database schema / what is stored:

1. I've noticed that "skipped" is also used if a particular question is skipped; which is interesting, but I think we'll now that as `response` is NULL. I would *only* use this when the user pressed `skipped or postponed`
2. There is also a column `scale` created; however, given that it is only used for the LikertScale question type and as we're storing it in `responseOptions` anyways; I suggest to delete the column. what do you think?

(note that the comment is not in the right place)

src/electron/src/views/DailySurveyView.vue
```diff
@@ -0,0 +1,566 @@
...
+<template>
+  <div ref="rootEl" class="daily-survey flex flex-col">
+    <div class="survey-header">
+      <div class="survey-title">
+        {{ samplingType === 'morning' ? 'Start-of-Workday' : 'End-of-Workday' }} Questionnaire
+      </div>
+    </div>
+
+    <div class="postpone-bar">
+      <button class="postpone-button" :disabled="isSubmitting" @click="postpone(5)">Remind me again in 5 minutes</button>
```
casaout, line 134: Postponed pop-ups don't show up again for me (tested with 5mins) interval.

casaout (changes requested):

src/electron/src/views/DailySurveyView.vue
```diff
@@ -14,7 +14,10 @@ const surveyConfig = studyConfig.trackers.dailySurveyTracker?.surveys?.find(
   (s) => s.samplingType === samplingType
 );
 const questions: ExperienceSamplingQuestion[] = surveyConfig?.questions ?? [];
-const requireAllAnswers = surveyConfig?.requireAllAnswers ?? false;
+const requireAllAnswers =
```
casaout: i don't understand how `requireAllAnswers` is set. what is the difference in this context between surveyConfig and studyConfig?

casaout (changes requested):

src/electron/src/views/DailySurveyView.vue
grigor-dochev: That's a good point! `studyConfig` is the full global study configuration and `surveyConfig` is the selected daily survey config for the current `samplingType` (`morning` or `evening`).

I originally had a per-survey override with a tracker-level fallback, but for now I can simplify it so `requireAllAnswers` is only configured once at the `dailySurveyTracker` level and applies to all daily surveys.

casaout: didn't we decide to change that to have one config per daily survey? (i.e. on survey level, NOT globally)

grigor-dochev: I don't remember, but I can definitely change it to be survey level only.

src/electron/electron/main/services/trackers/DailySurveyTracker.ts
casaout: thanks for making the change.

Just to clarify; what's the behavior of the daily survey when enabledWorkHours is set to false? -> will it consider the default work hours (not the user-enabled ones)?

grigor-dochev: Yes, exactly, that's the intended behaviour!

PR #518, opened 2026-04-28
[#386] Fix clean install onboarding startup screen

### LINKED ISSUE #386: Issues with
With #375, we've introduced additional locations where we're showing that the app is running (fine).

However, there are two bugs that we've discovered:

- [ ] on macOS, showing the wrong screen (discovered by @royru and [reported here](https://github.com/HASEL-UZH/PersonalAnalytics/pull/375#issuecomment-2634040367))
- [ ] on Windows, upon clean install, the last page instead of the entire onboarding is shown
- [ ] on macOS, verify that the entire onboarding screen and not just the last page is shown upon clean install (@royru)

grigor-dochev:
## Summary

- Fix clean-install onboarding startup so the full onboarding opens instead of the final "PersonalAnalytics is running" page
- Normalise persisted onboarding flags before branching on startup state
- Make macOS permission gating respect only permissions required by the study config

Closes #386

### DIFF
--- a/src/electron/electron/main/index.ts
+++ b/src/electron/electron/main/index.ts
@@ -137,15 +137,18 @@ app.whenReady().then(async () => {
     }
 
     const settings: Settings = await Settings.findOneBy({ onlyOneEntityShouldExist: 1 });
+    const onboardingShown = Boolean(settings.onboardingShown);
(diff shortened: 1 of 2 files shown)


royru (approved):
LGTM! Thanks Grigor!

PR #522, opened 2026-05-11
[#515] Add top websites and window titles

### LINKED ISSUE #515: Retrospection: Create Top Website/Window title
In the legacy retrospection, we have a table showing details on time spent in Files and Websites:

Image

We'd like to create two new cards to visualize (1) "Top websites" and (2) "Top window titles":
- Similar to "Top Activities" (#512), we'd like to visualize them visually appealing (i.e. in a small bar chart)
- I suggest to visualize the top 3 elements
- Websites you can extract from the category "Work related browsing" (no need to show the browser name)
- Top window titles from the remaining categories that are not browsing (no need to show the app name I think)

Some inspiration (but let's visualize it more like #512):

Image

These are ideas, happy to hear your thoughts and insights

grigor-dochev:
# Add Top Websites and Window Titles to Retrospection (Closes #515)

## Description

Adds two new Retrospection detail cards that summarize where participants spent the most time during the selected day: top websites and top window titles. The cards use compact horizontal bars, durations, and relative percentages to make the details easy to scan alongside the existing Retrospection timeline and insights.

### Changes Made

- New `Top websites` card showing the top 3 website domains for the selected day
- New `Top window titles` card showing the top 3 non-browsing window titles for the selected day
- Website entries are derived from `WorkRelatedBrowsing` activity and grouped by domain, without showing the browser app name
- Window title entries exclude browsing categories to avoid duplicating website activity
- `Other`, `Unknown`, and `Idle` entries are filtered out before selecting the top 3
- Relative percentage shown next to each duration, calculated within each card
- Card accent colors reuse the existing Retrospection activity color mapping used by the timeline legend
- Added IPC handlers and typed renderer commands for loading top websites and top window titles
- Updated Retrospection session aggregation to support derived grouping keys such as website domains and cleaned window titles
- Removed noisy service warnings for intentionally skipped rows during derived aggregation

## Implementation Details

1. Retrospection session reconstruction now supports grouping by a derived key, not only direct fields such as `processName` or `activity`. This allows website domains to be extracted from URLs and window titles to be cleaned before aggregation.
2. `getTopWebsiteSessions` only considers `WorkRelatedBrowsing` entries, extracts the hostname from each URL, groups active time by domain, filters irrelevant values, sorts by total duration, and returns the top 3.
3. `getTopWindowTitleSessions` excludes browsing categories (`WorkRelatedBrowsing`, `WorkUnrelatedBrowsing`, and `SocialMedia`), strips matching app suffixes from window titles, groups active time by cleaned title, filters irrelevant values, sorts by total duration, and returns the top 3.
4. Each returned top item carries its source activity so the frontend can use the same category color mapping as the timeline.
5. Rows with no derived key are treated as expected separators during aggregation instead of creating log warnings.

## Screenshots

| Light mode | Dark mode |
|:-:|:-:|
| Retrospection top websites and window titles in light mode | Retrospection top websites and window titles in dark mode |

Closes #515


src/electron/src/views/RetrospectionView.vue
```diff
@@ -229,15 +360,102 @@ function getDayLabel(date: Date): string {
...
+                {{ ACTIVITY_LABELS[activitySession.type] || 'Other' }}:
+                {{ renderTime(activitySession.totalDurationMs) }}
               </li>
             </ol>
           </div>
+        </div>
 
+        <!-- Website and window title details -->
+        <div v-if="topItemsAvailable">
+          <div class="mb-2 mt-8">
```
casaout: i don't think we'll need another title

also in terms of what is displayed, it's very similar to "top apps" and "top activities pursued" anyways

this will also help to reduce the window size

src/electron/src/views/RetrospectionView.vue
```diff
@@ -229,15 +360,102 @@ function getDayLabel(date: Date): string {
...
+            <h1 class="text-xl font-bold text-gray-900 dark:text-gray-100">Details of your day</h1>
+          </div>
+          <div class="tile-grid">
+            <div
+              v-if="topWebsites.length"
+              class="top-item-card rounded border border-gray-200 bg-gray-100 px-4 py-3 text-gray-800 dark:border-transparent dark:bg-neutral-800 dark:text-slate-200"
+            >
+              <h2 class="primary-blue font-bold leading-4">Top websites</h2>
```
casaout: note that I don't see the "top website" box (tested on multiple days), even though "Microsoft Edge use" and "Work Related Browsing" was quite big overall

Image

(so i couldn't test that feature yet)

src/electron/src/views/RetrospectionView.vue
```diff
@@ -229,15 +360,102 @@ function getDayLabel(date: Date): string {
...
+                      class="top-item-label"
+                      :title="`${website.type} · ${getTopItemActivityLabel(website)}`"
+                      >{{ website.type }}</span
+                    >
+                    <span class="top-item-metric">
+                      <span class="top-item-time">{{
+                        renderCompactTime(website.totalDurationMs)
+                      }}</span>
+                      <span class="top-item-percentage">{{
+                        getTopItemPercentage(website, topWebsites)
```
casaout: I think we should try to align the design slightly more to the existing design; few suggestions:

1. the numbering can be simplified (no rounded border); or maybe even removed as it'll be intuitive which one is the highest one?
2. the fonts should be the same as for the other insight boxes
3. in the dark mode, i think the light background of the chart is a bit too strong
4. i think we don't need to display the percentages (as the timeline chart will imply this anyways), and to have more space for the window title)
5. it can be made a bit more compact, to reduce the overall height of the insights

Great that you're using the activity colors; that's awesome!!

src/electron/src/views/RetrospectionView.vue
```diff
@@ -229,15 +360,102 @@ function getDayLabel(date: Date): string {
...
+                    <span
+                      class="top-item-label"
+                      :title="`${website.type} · ${getTopItemActivityLabel(website)}`"
+                      >{{ website.type }}</span
+                    >
+                    <span class="top-item-metric">
+                      <span class="top-item-time">{{
+                        renderCompactTime(website.totalDurationMs)
+                      }}</span>
+                      <span class="top-item-percentage">{{
```
casaout: once we finalize on the design of these "charts", i think we can uniformly use the same design (and CSS) for all 4 boxes: top apps/activities & websites

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +93,230 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+ * @returns all usage sessions per prop, including the total duration
+ */
+async function getWindowActivitySessionsByType(
+  prop: 'processName' | 'activity',
+  date: Date
+): Promise<ActivitySessions[]> {
+  return await getWindowActivitySessionsByKey((activity) => activity[prop], date);
+}
+
+function getDomainFromUrl(url: string | null): string | null {
```
casaout, line 375: guiding from your screenshot, i think just having the domain is not sufficient to know what a person was doing; e.g. on github it could be specification, code review, browsing, etc

is it possible to show slightly more (i.e. the core part of the displayed window title)?

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +93,230 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+    return null;
+  }
+
+  let title = windowTitle.trim();
+  if (!title) {
+    return null;
+  }
+
+  if (processName) {
+    const suffixes = [` - ${processName}`, ` — ${processName}`, ` | ${processName}`];
```
casaout: i think one more cleaning step should be in case a path is part of the window title; see example:

Image

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +93,230 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+    const matchingSuffix = suffixes.find((suffix) => title.endsWith(suffix));
```
casaout: When matching an URL to an activity (not website); e.g. "Overleaf" to "read/write doucment", i think the window/browser cleaning doesn't fully work yet. it's great that it shows the website name first; but on hover one sees for example "4 or more pages"

Image

casaout (changes requested):
great start, thank you!

I now mostly reviewed UI; and will review the code once we've settled on the design

src/electron/src/views/RetrospectionView.vue
grigor-dochev: Screenshot 2026-05-18 at 14 19 55

casaout: It already looks better, thank you. However, to me still looks a bit like it's in light mode (with the bright colors)...
- I wonder if it'd work better if the barchart was the actual category color 
- the font was white (in dark mode) and dark (as is in light mode)? (then it'd look much more similar to the activities timeline (which also helps to clarify the colors)
- And the "white" part could be white in light mode; and a very dark shade (i.e. the background color); so the white font would easily be readable

Also, can we vertically make it slightly more compact, such as by removing some of the top/bottom padding between the text and the barchart?

grigor-dochev: > It already looks better, thank you. However, to me still looks a bit like it's in light mode (with the bright colors)...
> 
> * I wonder if it'd work better if the barchart was the actual category color
> * the font was white (in dark mode) and dark (as is in light mode)? (then it'd look much more similar to the activities timeline (which also helps to clarify the colors)
> * And the "white" part could be white in light mode; and a very dark shade (i.e. the background color); so the white font would easily be readable
> 
> Also, can we vertically make it slightly more compact, such as by removing some of the top/bottom padding between the text and the barchart?

Screenshot 2026-05-19 at 11 31 37

What do you think of this?

casaout: that's great, thanks! Do we need the broders at all? and how dos it look in light mode? thanks!

src/electron/src/views/RetrospectionView.vue
```diff
@@ -229,15 +352,78 @@ function getDayLabel(date: Date): string {
...
               <li v-for="(activitySession, index) in topActivities" :key="index">
-                {{ ACTIVITY_LABELS[activitySession.type] || 'Other' }}: {{ renderTime(activitySession.totalDurationMs) }}
+                {{ ACTIVITY_LABELS[activitySession.type] || 'Other' }}:
+                {{ renderTime(activitySession.totalDurationMs) }}
               </li>
             </ol>
           </div>
+        </div>
 
+        <div v-if="topItemsAvailable" class="tile-grid mt-8">
```
casaout: spacing between the old rows and new rows is off (see red arrows):

Image

(please make it lower for the new row, same as for the previous ones)

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -1,28 +1,42 @@
...
-  sessions: TimeActive[]
+  type: string;
+  totalDurationMs: number;
+  sessions: TimeActive[];
+  activity?: string;
 }
 
+type WindowActivitySessionKeySelector = (activity: WindowActivityEntity) => string | null;
+
+const BROWSER_PROCESS_NAMES = [
```
casaout: what do we need this list for? (please add a quick comment to remember this later)

also, i wonder if we could reuse the list from https://github.com/HASEL-UZH/PA.WindowsActivityTracker/blob/main/typescript/src/mappings/browsers.ts and then compare the process name instead of app name? (and also extend it with the list from there)

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +104,325 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+  const session = { from, to, duration };
+  if (activity && !entry.activity) {
+    entry.activity = activity;
+  }
+  entry.sessions.push(session);
+  entry.totalDurationMs += duration;
+  map.set(key, entry);
+}
+
+function addSessionWithActiveMinuteSplits(
```
casaout, line 230: can you briefly explain (and document) the goal and logic of this method?

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +104,325 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
- * active window activity sessions for apps or activities (e.g., "Planning", "InstantMessaging", ...)
- * @param prop - the property to group by, either "processName" for apps or "activity" for activities
+ * active window activity sessions grouped by a derived key.
+ * @param getSessionKey - derives the session grouping key from a raw window activity entry
  * @returns all usage sessions per prop, including the total duration
  */
-async function getWindowActivitySessionsByType(prop: "processName" | "activity", date: Date): Promise<ActivitySessions[]> {
-  const windowActivityToday = await getWindowActivities(date)
-  const activeMinutesSet = await getActiveMinutesSet(date)
+async function getWindowActivitySessionsByKey(
```
casaout, line 295: can you please explain this method a bit; and also what changed from the existing one?

what is a sessionkey?

src/electron/electron/main/services/RetrospectionService.ts
casaout: thanks for the updates, however,w ithout in-code documentation, it is very difficult to understand what exactly is done.

I suggest to make an example and describe the supsteps

casaout (commented):
thanks for the updates!  To understand the "algorithm" for loading the actual datasets, extracting window titles, I think it'd be beneficial to add more documentation *and* rationale. otherwise, it's difficult to review the code, and also later maintain it. Thanks!

src/electron/electron/main/services/RetrospectionService.ts
grigor-dochev: The tracker doesn't export the list a sa runtime import and it's being used to recognise the browser process names for title cleanup. So, for now, I've mirred the tracker browser mapping instead.

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +139,395 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+    /^(new tab|about:blank|start page|untitled)$/i.test(title)
+  );
+}
+
+/**
+ * Turns URL-like window titles into a compact, readable label.
+ *
+ * Some browsers expose a title that is effectively the URL, for example
+ * "github.com/HASEL-UZH/PersonalAnalytics/pull/123". For display we keep the domain plus the last
+ * two path parts, producing "github.com/pull/123". This avoids unreadable full paths while still
```
casaout: i like that solution a lot! I was, however just wondering, if it should be github.com/.../pull/123 (i.e. add "..." to indicate it's shortened)

and maybe the same for stripPathFragment?

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +139,395 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+ * Substeps:
+ * 1. Trim empty titles.
+ * 2. Split on common title separators (`-`, `—`, `–`, `|`).
+ * 3. Shorten path-like fragments and URL-like fragments.
+ * 4. Remove trailing app/browser suffixes such as "Microsoft Edge".
+ * 5. Remove generic browser prefixes such as "4 or more pages".
+ * 6. If the remaining title is still generic, fall back to the captured URL domain.
+ *
+ * Examples:
+ * "4 or more pages - Overleaf - Microsoft Edge" -> "Overleaf"
```
casaout, line 544: this might not work in all cases yet (as discussed in our meeting):

Image

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -95,125 +139,395 @@ export async function getWindowActivities(date: Date | string): Promise<WindowAc
...
+    .filter(isRelevantTopItem)
+    .sort((a, b) => b.totalDurationMs - a.totalDurationMs)
+    .slice(0, limit);
+}
+
+export async function getTopWindowTitleSessions(
+  date: Date,
+  limit = 3
+): Promise<ActivitySessions[]> {
+  const browsingActivities = new Set([
```
casaout: as discussed in the meeting, please also include Coding, Read/Write Document and Planning

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
casaout: Sorry for not being clear earlier, but I think we also don't need the background color; i.e. just the window/website titel (left aligned with the card's title), the bar and the time

image

src/electron/electron/main/services/RetrospectionService.ts
casaout: this is still happening:
image

src/electron/electron/main/services/RetrospectionService.ts
casaout: This is also still happening; to clarify, i think here it shoudl show something like C...DevEx in practice: (i.e. add "..." whenever a filepath or website path are shortened)

image

casaout (changes requested):

src/electron/electron/main/services/RetrospectionService.ts
casaout: this is still not solved...

src/electron/electron/main/services/RetrospectionService.ts
casaout: better, but the *and* is still in there:

image

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -86,6 +87,15 @@ const EXCLUDED_TOP_WINDOW_TITLE_ACTIVITIES = new Set([
   'SocialMedia'
 ]);
 
+const ACTIVITY_TITLE_SUFFIX_NAMES = [
```
casaout: how is this list used; and wouldn't it be better to match for the activities-enum/type (instead of the names that might change)?

src/electron/scripts/test-retrospection-title-utils.cjs
```diff
@@ -0,0 +1,208 @@
+const assert = require('node:assert/strict');
```
casaout: i think we should create a separate folder for tests, like we did for the trackers; e.g. https://github.com/HASEL-UZH/PA.WindowsActivityTracker/tree/main/typescript/src/__tests__

src/electron/package.json
```diff
@@ -19,7 +19,8 @@
     "preview": "vite preview",
     "lint": "eslint . --ext .js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts,.vue --fix",
     "lint:fix": "eslint --fix . --ext .js,.jsx,.ts,.tsx,.vue",
-    "format": "prettier --write ."
+    "format": "prettier --write .",
+    "test:retrospection": "node scripts/test-retrospection-title-utils.cjs"
```
casaout: please use the same style of testing as in the other PA-repos: https://github.com/HASEL-UZH/PA.WindowsActivityTracker/blob/7121a19dbb0856d7487e9b0b0a17696a26bb8b53/typescript/package.json#L18

src/electron/scripts/test-retrospection-title-utils.cjs
```diff
@@ -0,0 +1,243 @@
...
+      'C:/Users/<username>/OneDrive//Research/Project/2026-Project.mqda - MAXQDA Analytics Pro (26.2.1)',
+      'MAXQDA Analytics Pro'
+    ),
+    expected: '2026-Project.mqda'
+  },
+  {
+    name: 'Edge grouped-tabs title strips tab-count fragment',
+    actual: cleanWindowTitle(
+      'Structure - Chapters & Plays - Google Docs and 2 more pages - School - Microsoft​ Edge',
+      'Microsoft Edge',
```
casaout: can you maybe also add a few testcases of the browser of your choice (Chrome/Safari), so we have a broader set of cases where the title adds whings such as "new tab", "untitled", "and more pages"?

casaout (changes requested):

casaout (approved):

casaout (approved):

PR #523, opened 2026-05-11
[#514] Add activity breakdown pie chart to Retrospection

### LINKED ISSUE #514: Retrospection: Add pie chart for Activities
Similar to the legacy PersonalAnalytics, create a new card "Activity Breakdown" showing the top activities (accounting for e.g. 90% of the total time) as a small pie chart. 

Image

It's okay to drop the minor ones, e.g. for the following spread, I think it'd be enough to have Coding, Work related browsing, GAI, email, IM (and maybe Planning):

Image

Let's make it compact, inspiration:

Image

grigor-dochev:
# Add Activity Breakdown to Retrospection (Closes #514)

## Description

Adds a compact `Activity Breakdown` card to Retrospection. The card shows the top activity categories for the selected day as a donut chart with durations.

### Changes Made

- Added `Activity Breakdown` card
- Groups activities using the existing Retrospection activity labels and colors
- Shows the largest activity categories covering roughly 90% of tracked time
- 
## Screenshots

| Light mode | Dark mode |
|:-:|:-:|
| Activity breakdown in light mode | Activity breakdown in dark mode |

Closes #514

### DIFF
--- a/src/electron/src/views/RetrospectionView.vue
+++ b/src/electron/src/views/RetrospectionView.vue
@@ -1,54 +1,146 @@
 <script setup lang="ts">
-import { computed, onMounted, ref } from 'vue'
-import { Activity, ActivitySessions, ChartDataPoint, DataPointType, TimeActive } from '../utils/retrospection/types'
-import { ACTIVITY_LABELS, getTailwindClassFromActivity } from '../utils/retrospection/utils'
-import StackedBarChart from '../components/StackedBarChart.vue'
+import { computed, onMounted, ref, type CSSProperties } from 'vue';
+import {
+  Activity,
+  Color,
+  DataPointType,
+  type ActivitySessions,
+  type ChartDataPoint,
+  type PieChartDataPoint,
+  type TimeActive
+} from '../utils/retrospection/types';
+import {
+  ACTIVITY_LABELS,
+  getActivityGroupFromActivityName,
+  getTailwindClassFromActivity
+} from '../utils/retrospection/utils';
+import StackedBarChart from '../components/StackedBarChart.vue';
 
-// @ts-ignore
-const typedIpcRenderer = window.ipcRenderer
-const isLoading = ref(false)
-const selectedDay = ref(new Date())
-const allWindowActivities = ref<ActivitySessions[]>([])
-const chartDataWindowActivities = ref<ChartDataPoint[]>()
-const longestTimeActive = ref<TimeActive | undefined>(undefined)
-const topApps = ref<ActivitySessions[] | undefined>(undefined)
-const topActivities = ref<ActivitySessions[]>([])
+const typedIpcRenderer = window.ipcRenderer;
+const isLoading = ref(false);
+const selectedDay = ref(new Date());
+const allWindowActivities = ref<ActivitySessions[]>([]);
+const chartDataWindowActivities = ref<ChartDataPoint[]>();
+const longestTimeActive = ref<TimeActive | undefined>(undefined);
+const topApps = ref<ActivitySessions[] | undefined>(undefined);
+const topActivities = ref<ActivitySessions[]>([]);
+const ACTIVITY_BREAKDOWN_COVERAGE = 0.9;
+
+interface ActivityBreakdownDataPoint extends PieChartDataPoint {
+  percentage: number;
+}
 
 const earliestUserComputerActivity = computed((): number => {
-  return chartDataWindowActivities.value?.reduce((acc, activity) => {
-    if (activity.start.getTime() < acc) {
-      return activity.start.getTime()
-    }
-    return acc
-  }, Number.MAX_SAFE_INTEGER) ?? Number.MAX_SAFE_INTEGER
-})
+  return (
+    chartDataWindowActivities.value?.reduce((acc, activity) => {
+      if (activity.start.getTime() < acc) {
+        return activity.start.getTime();
+      }
+      return acc;
+    }, Number.MAX_SAFE_INTEGER) ?? Number.MAX_SAFE_INTEGER
+  );
+});
 
 const latestUserComputerActivity = computed((): number => {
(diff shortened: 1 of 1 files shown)


src/electron/src/views/RetrospectionView.vue
```diff
@@ -265,4 +453,108 @@ h2.primary-blue {
...
+}
+
+.activity-breakdown-dot {
+  width: 0.65rem;
+  height: 0.65rem;
+  border-radius: 50%;
+  flex: 0 0 auto;
+}
+
+.activity-breakdown-name {
```
casaout, line 529: can you please align the font with the one for the other texts/boxes?

src/electron/src/views/RetrospectionView.vue
```diff
@@ -58,186 +150,282 @@ function windowActivitiesToChartData() {
...
+                  :title="`${activity.name}: ${renderCompactTime(activity.value)}`"
+                >
+                  <span class="activity-breakdown-label">
+                    <span
+                      class="activity-breakdown-dot"
+                      :style="{ backgroundColor: activity.color }"
+                    ></span>
+                    <span class="activity-breakdown-name">{{ activity.name }}</span>
+                  </span>
+                  <span class="activity-breakdown-time">{{
```
casaout: having the time is great, but in dark mode, it's not super readable. maybe make it slightly lighter?

Image

(in light mode it looks great)

src/electron/src/views/RetrospectionView.vue
```diff
@@ -58,186 +150,282 @@ function windowActivitiesToChartData() {
...
+            class="activity-breakdown-card rounded border border-gray-200 bg-gray-100 px-4 py-3 text-gray-800 dark:border-transparent dark:bg-neutral-800 dark:text-slate-200"
+          >
+            <h2 class="primary-blue font-bold leading-4">Activity Breakdown</h2>
+            <div class="activity-breakdown-content">
+              <div class="activity-pie" :style="activityBreakdownStyle" aria-hidden="true">
+                <div class="activity-pie-hole"></div>
+              </div>
+              <ol class="activity-breakdown-list">
+                <li
+                  v-for="activity in activityBreakdownData"
```
casaout, line 413: to avoid making it too big when having many different activity categories, i suggest to only show the top X:

Image

I think I suggested to consider only the top 90% of activities; and in this case (as most have a similar distribution) this would correctly use more of them; but maybe we still cut the list at 6 items? and in that case show in the header ("Activity Breakdown (top 6 activities)")?

src/electron/src/views/RetrospectionView.vue
```diff
@@ -1,54 +1,146 @@
...
+      }
+      if (activity.end.getTime() > acc) {
+        return activity.end.getTime();
+      }
+      return acc;
+    }, 0) ?? 0
+  );
+});
+
+const activityBreakdownTotalMs = computed((): number => {
```
casaout, line 60: for each of these functions, can you please briefly summarize in an in-line code comment what they do (I think that this is not always easily graspable from the code itself) and will hopefully aid maintenance

casaout (changes requested):
Looks great!

src/electron/src/views/RetrospectionView.vue
```diff
@@ -58,186 +150,282 @@ function windowActivitiesToChartData() {
...
             <ol class="mt-2 list-decimal pl-4">
               <li v-for="(appSession, index) in topApps" :key="index">
                 {{ appSession.type }}: {{ renderTime(appSession.totalDurationMs) }}
               </li>
             </ol>
           </div>
 
-          <!-- Tile 4: Top activities -->
-          <div v-if="topActivities" class="text-gray-800 bg-gray-100 border border-gray-200 rounded px-4 py-3 dark:text-slate-200 dark:bg-neutral-800 dark:border-transparent">
-            <h2 class="leading-4 primary-blue font-bold">Top activities pursued</h2>
```
casaout, line 233: given that the new "activity breakdown" shows almost the same information as "top activities pursued", i think we can now remove the "top activities pusued" insight box

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
casaout: image

It looks great, but 
- can you reduce the spacing between the legend items (to decrease the overall height of the box)
- also, the times shouldn't be in bold (I know they are for the boxes longest active times and active hours on computer; but that's because they're in a text)

src/electron/src/views/RetrospectionView.vue
```diff
@@ -1,54 +1,160 @@
...
+  allWindowActivities.value.forEach((activitySession) => {
+    const activityGroup = getActivityGroupFromActivityName(activitySession.type);
+    groupTotals.set(
+      activityGroup,
+      (groupTotals.get(activityGroup) ?? 0) + activitySession.totalDurationMs
+    );
+  });
+
+  const sortedActivities = Array.from(groupTotals.entries())
+    .filter(([activityGroup]) => activityGroup !== 'Other')
```
casaout: should we also filter "unknown"?

src/electron/src/views/RetrospectionView.vue
```diff
@@ -1,54 +1,160 @@
...
+    .map(([activityGroup, value]) => {
+      const colorKey = getTailwindClassFromActivity(activityGroup, true) as keyof typeof Color;
+      return {
+        name: ACTIVITY_LABELS[activityGroup] || 'Other',
```
casaout: isn't "other" already filtered out?

casaout (commented):
looks good; i just have three minor comments left. thanks!

src/electron/src/views/RetrospectionView.vue
grigor-dochev: That's true, this check is redundant.

Issue #527, opened 2026-05-21
Refactor & Refine Retrospection


casaout:
The `RetrospectionService` is already 500+ lines and still growing, as new figures are being added (#510, #511, #512, #513, #515). All data-fetching, transformation, and chart-config logic lives in a single file, which makes it:

- Hard to navigate and reason about
- Harder to test individual figures in isolation
- Likely to cause merge conflicts as multiple figures are worked on in parallel
- Increasingly difficult to onboard new contributors

## Tasks
feel free to split into sub-issues where useful; also as the list grew quite a bit; maybe making individual PRs makes sense

- [x] refactor retrospectionservice (see proposed solution below)
- [x] please use the design of #515 also for  "Top apps"
- [x] for the "time spent" bar charts, also show app/website icons (reuse code from the hover-detail info #510 )
- [x] let's discuss the order of the items to match box-height and if we want to increase the size of the window (and possibly add a third column)
- [ ] check the retrospection when one or several of the trackers are disabled (or no/not enough data is available)
- [x] optimize the size of the window (also increasing the width adjusts the boxes, but not the timeline)
- [x] for timeline views, unify the method for start/end timestamp (see [discussion here](https://github.com/HASEL-UZH/PersonalAnalytics/pull/540#discussion_r3512641672))
- [x] #555 
- [x] https://github.com/HASEL-UZH/PA.WindowsActivityTracker/issues/25
- [ ] https://github.com/HASEL-UZH/PA.SelfReflection/issues/24

## Proposed Solution for Refactoring

Extract each figure into its own self-contained module, fore xample:
retrospection/
figures/
useActiveHours.ts          // #513
useTopActivities.ts        // #512
useTimeline.ts             // #510
useSelfReports.ts          // #511
useTopWebsites.ts          // #515
RetrospectionService.ts      // thin orchestrator / registry
RetrospectionHelpers.ts

Each figure module would be responsible for its own (if that makes sense(
- DB query / data fetching
- Data transformation
- Chart/figure config (if applicable)

Issue #513, opened 2026-04-17
Retrospection: Refine "Active hours on computer"


casaout:
The current card that visualizes active hours on the computer might be a bit misleading:
- it visualizes the span of active work time (i.e. first to last entry for the day), not necessarily if the user was active

<img width="401" height="115" alt="Image" src="https://github.com/user-attachments/assets/b58b2e3d-f886-431b-ac4a-7af705d36df7" />

Thus, let's visualize both in that card:
- title: "Active hours"
- Active on the computer: X.x hours
- Spanning work: X.x hours (between 07:25 and 14:51)     <- number we have today

Issue #510, opened 2026-04-17
Retrospection: Refine Timeline


casaout:
When hovering over the timeline, one currently only sees the activity itself:

<img width="800" height="214" alt="Image" src="https://github.com/user-attachments/assets/8c6211d5-db62-4582-97ce-0808a2c0ecf6" />

In the legacy retrospection, we also showed the (most used) window titles & app names:

<img width="828" height="298" alt="Image" src="https://github.com/user-attachments/assets/268b4616-b5c4-40b9-abe1-f300c6780f34" />

We'd like to extend the hover overlay to provide more insights. As in some cases there might be dozens of websites, please consider what could be useful (e.g. only showing the top 3, trimming the titles (e.g. removing the app name and showing a logo, etc.)

After exploring options, we can discuss details.

Tasks:
- [ ] explore good ways
- [ ] implement after discussion
- [ ] also check what happens to retrospection if the windowsactivitytracker is disabled by the researcher in study.config.ts

Issue #22, opened 2016-02-18
Refactor TaskTrayIcon 


casaout:
to make the toast notifications & menu items accessible from the shared library

casaout:
Closing as reported for legacy version of PersonalAnalytics.

PR #540, opened 2026-06-19
[#511] Retrospection: Visualize Self-Reports

### LINKED ISSUE #511: Retrospection: Visualize Self-Reports
The legacy retrospection visualizes self-reports in a timeline view:

Image

With the new version, we need to be more flexible, as there might be likert-scale responses of different questions.
Also, it'd be nicer to visually align the self-reports on the time-axis with the timeline (if available).

Please explore how to best visualize self-reports over time. Few early ideas are:
- visualize self-reports in the same timeline span (i.e. start + end time) below the activity one as a linechart
- similar to enabling/disabling activities in the legend; show a legend if there are several questions around and let the user (un)select them
- i was thinking of overlays, but I think visibility might be difficult (especially if at some point we also visualize the user input as well)

Tasks:
- [ ] explore good ways
- [ ] implement after discussion
- [ ] also check what happens to retrospection if the experience sampling tracker is disabled by the researcher in study.config.ts or if no self-reports are available

grigor-dochev:
# Visualize Self-Reports in Retrospection (Closes #511)

## Description

Adds a `Self-reports over time` visualization to Retrospection. The chart shows Likert-scale self-report responses over time, aligned with the activity timeline when activity data is available.

### Changes Made

- Added day-scoped self-report loading for Retrospection
- Added a self-report line chart below the activity timeline
- Groups responses by self-report question
- Normalizes different Likert scales so multiple questions can be shown together
- Adds an activity-style legend for focusing/deselecting question series
- Keeps Retrospection available when self-reports exist but activity data is missing
- Hides the self-report section when the experience sampling tracker is disabled
- Shows an empty self-report state when no self-reports are available for the selected day

## Screenshots

| Default state | Legend item selected + tooltip | No self-reports |
|:-:|:-:|:-:|
| Self-reports over time default state | Self-reports over time with selected legend item and tooltip | Retrospection with no self-reports available |

Closes #511


src/electron/src/components/SelfReportTimelineChart.vue
```diff
@@ -0,0 +1,388 @@
...
+    if (Array.isArray(parsed)) {
+      return parsed;
+    }
+  } catch {
+    return [];
+  }
+  return [];
+}
+
+function getLabel(point: SelfReportPoint, position: 'low' | 'mid' | 'high'): string {
```
casaout, line 174: how are the labels in the y-axis defined? (currently, i see high/mid/low, but I am assuming that this is not always the right label...)

few thoughts (i don't have a great solution yet, so something to continue discussing):
- when selecting a single question, we can show the actual labels
- when all questions have the same scale, we could also show the actual labels
- when question have different labels, maybe we show the numbers (1-7 or 1-5)

Image

src/electron/src/views/RetrospectionView.vue
```diff
@@ -435,35 +552,69 @@ function getDayLabel(date: Date): string {
...
+          <StackedBarChart
+            v-if="!isLoading && chartDataWindowActivities"
+            :data="chartDataWindowActivities"
+            :start-date="getNearestFullHourTime(earliestUserComputerActivity, 0)"
+            :end-date="getNearestFullHourTime(latestUserComputerActivity, 1)"
+            type="WINDOW_ACTIVITY"
+          />
+        </template>
+
+        <template v-if="experienceSamplingEnabled">
```
casaout: few minor comments:

- note that a comment is missing (there is one for the timeline and info tiles)
- i think the self-reports visualization should come after the boxes (as i think they are more actionable/useful)
- also the vertical height of the chart can be reduced
- can the "points" indicating the value maybe be a bit bigger and the lines dotted/dashed to imply that it's a momentary value and they just indicate rough trends?


(generally, i guess the chart will become more useful over longer durations (not single day) and when connecting it to other data; but that's for later I guess)

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
```diff
@@ -394,7 +487,18 @@ function getDayLabel(date: Date): string {
...
+            xmlns="http://www.w3.org/2000/svg"
+            viewBox="0 0 24 24"
+            fill="none"
+            stroke="currentColor"
+            stroke-width="2.5"
+            stroke-linecap="round"
+            stroke-linejoin="round"
+            class="h-4 w-4"
+          >
+            <polyline points="9 18 15 12 9 6" />
```
casaout, line 509: the hover info is good to have, but it's a bit inconsistent, i.e. once it shows the label and once not:

Image

Image

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
grigor-dochev: Image

casaout: it's better, but I think still a bit hard to parse; few ideas:

- left align the text

- Rating: 5 on a scale from 1-7  (not stressed / ...)
- Self-report taken at: 10:25

src/electron/src/components/SelfReportTimelineChart.vue
casaout: wroks well; the only change i would make is that when hovering over a question, to show the entire questioN (useful if it's a long one that is cut off):

Image

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
grigor-dochev: Screenshot 2026-06-29 at 15 45 04

src/electron/src/views/RetrospectionView.vue
```diff
@@ -62,6 +67,64 @@ const topItemsAvailable = computed((): boolean => {
...
+});
+
+const latestSelfReport = computed((): number => {
+  return selfReports.value.reduce((acc, report) => {
+    const promptedAt = new Date(report.promptedAt).getTime();
+    return promptedAt > acc ? promptedAt : acc;
+  }, 0);
+});
+
+const timelineStartDate = computed((): number => {
```
casaout, line 105: just to clarify: as this and timelineEndDate is likely also used for other timelines (but was coded in separate parallel issues), is the same function (re)used?

casaout (approved):

src/electron/src/views/RetrospectionView.vue
grigor-dochev: I checked the parallel branches (#510, #513, and AIRBar #22). There doesn't seem to be a shared helper for this yet; the timeline bounds are still computed locally in `RetrospectionView.vue` and then passed into the respective timeline/chart components. Here, I reused the same computed bounds for the self-report chart so it aligns with the activity timeline.

casaout: Sorry if I wasn't clear. what i meant was that for the other timelines, don't we also need the start and end timestamp? Then, we could refactor this (maybe as part of #527) to use the same helper method. but before updating the issue, i wanted to ask if this makes sense

grigor-dochev: Ok I understand now! Yes, that makes sense. Other timelines also need the same start/end timestamps, so a shared helper would be useful. It makes sense to handle it in #527, once the parallel branches are merged.

casaout (approved):

PR #544, opened 2026-06-23
[#510] Refine retrospection timeline hover details

### LINKED ISSUE #510: Retrospection: Refine Timeline

grigor-dochev:
# Refine Retrospection Timeline Hover Details (Closes #510)

## Description

Improves the `Activities over time` hover overlay in Retrospection. The tooltip now gives more context for each activity segment by showing the most relevant window titles, app names, durations, and app icons.

### Changes Made

- Added timeline hover detail aggregation for activity sessions
- Shows the top window/app details for each hovered activity segment
- Cleans and trims noisy window titles before display
- Removes redundant app/title suffixes where possible
- Adds native app icons from captured process paths
- Falls back to compact initial badges when an app icon is unavailable
- Keeps tooltip content bounded to the top items with a `+ N more` row
- Repositions the tooltip above the timeline segment
- Clamps tooltip placement so it does not overflow outside the viewport
- Adds a clearer Retrospection empty state when window activity tracking is disabled

## Screenshots

| Timeline hover with app/window details | Timeline hover with multiple details |
|:-:|:-:|
| Retrospection timeline hover with app and window details | Retrospection timeline hover showing multiple window details |

Closes #510

### DIFF
--- a/src/electron/electron/main/__tests__/RetrospectionService.test.ts
+++ b/src/electron/electron/main/__tests__/RetrospectionService.test.ts
@@ -1,6 +1,17 @@
 import { expect, jest, test } from '@jest/globals';
 
 jest.unstable_mockModule('electron', () => ({
+  app: {
+    getFileIcon: jest.fn()
+  },
+  nativeImage: {
+    createFromPath: jest.fn(() => ({
+      isEmpty: () => true,
+      resize: () => ({
+        toDataURL: () => ''
+      })
+    }))
+  },
   default: {
     app: {
       isPackaged: false
--- a/src/electron/electron/main/services/RetrospectionService.ts
+++ b/src/electron/electron/main/services/RetrospectionService.ts
@@ -3,6 +3,7 @@ import { WindowActivityEntity } from '../entities/WindowActivityEntity';
 import { getMainLogger } from '../../config/Logger';
 import { Activity } from '../../../src/utils/retrospection/types';
 import { getProcessIconDataUrl } from './utils/AppIconHelper';
+import { getOverlapDurationMs } from './utils/helpers';
 
 const LOG = getMainLogger('RetrospectionService');
 
@@ -689,8 +690,10 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
   );
 }
 
-// Short display label shown inside the tooltip row; keeps paths/domains compact.
-function getTimelineHoverTitle(activity: WindowActivityEntity): string | null {
+/**
+ * Returns the short display label shown inside a timeline tooltip row.
+ */
+function getShortTimelineHoverTitle(activity: WindowActivityEntity): string | null {
   const cleanedTitle = cleanWindowTitle(
     activity.windowTitle,
     activity.processName,
@@ -703,8 +706,10 @@ function getTimelineHoverTitle(activity: WindowActivityEntity): string | null {
   );
 }
 
-// Longer hover title for the row's native browser tooltip; keeps ellipses for shortened paths.
-function getTimelineHoverTooltipTitle(activity: WindowActivityEntity, title: string): string {
+/**
+ * Returns the longer native tooltip label for a timeline tooltip row.
+ */
+function getLongTimelineHoverTitle(activity: WindowActivityEntity, fallbackTitle: string): string {
   return (
     cleanWindowTitle(
       activity.windowTitle,
@@ -712,7 +717,7 @@ function getTimelineHoverTooltipTitle(activity: WindowActivityEntity, title: str
       activity.url,
       true,
       activity.activity
-    ) || title
+    ) || fallbackTitle
   );
 }
 
@@ -724,12 +729,12 @@ async function addWindowActivityDetailSpan(
   activeMinutesSet: Set<number>,
   date: Date
 ) {
-  const title = getTimelineHoverTitle(activity);
+  const title = getShortTimelineHoverTitle(activity);
   if (!title || to.getTime() <= from.getTime()) {
     return;
   }
 
-  const tooltipTitle = getTimelineHoverTooltipTitle(activity, title);
+  const tooltipTitle = getLongTimelineHoverTitle(activity, title);
   const iconDataUrl = await getProcessIconDataUrl(activity.processPath, activity.processName);
   getActiveMinuteSpans(from, to, activeMinutesSet, date).forEach((span) => {
     spans.push({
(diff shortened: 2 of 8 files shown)


src/electron/electron/main/__tests__/RetrospectionService.test.ts
```diff
@@ -1,6 +1,17 @@
 import { expect, jest, test } from '@jest/globals';
 
 jest.unstable_mockModule('electron', () => ({
```
casaout, line 3: what does this change to the test file do?

src/electron/src/views/RetrospectionView.vue
```diff
@@ -374,12 +378,25 @@ function getDayLabel(date: Date): string {
...
-        >Today</button>
+        >
+          Today
+        </button>
         <button
           class="h-8 rounded border border-gray-300 bg-white px-2 text-gray-800 dark:border-neutral-600 dark:bg-neutral-700 dark:text-slate-200"
           @click="navigateToPreviousDay"
         >
-          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" stroke-linejoin="round" class="h-4 w-4"><polyline points="15 18 9 12 15 6"/></svg>
+          <svg
```
casaout, line 388: It's a really great start!

A few suggestions for refinement are:
- i would hide items that are shorter than 1 min (i.e. < 00:01)
- also, i think we should show a unit; in addition as people usually spend only a few minutes in a file/website/app, i would probably show the time in minutes if it's less than 1 hour, and then "X h y min" if it's longer (note you're doing that already for the "top window titles", so you could reuse these methods)
- given the width of the hover window, i am wondering if the "from to" instead of on a second line, could be moved to the same line behind the activity category; and the "saved" line can be used to show the top 4 artifacts in the activity

Image

casaout (changes requested):

src/electron/src/views/RetrospectionView.vue
casaout: - to increase the amount of text we can read, can you please decrease the font size slightly?
- also, i think the time + duration (09:52 - 10:07 (16min)) should be in the default font (color)


image

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -637,6 +693,331 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
   );
 }
 
+function getTimelineHoverTitle(activity: WindowActivityEntity): string | null {
```
casaout: can you please add comments to the functions; to help understand the difference, e.g. between getTimelineHoverTitle and getTimelineHOverTooltipTitle

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -637,6 +693,331 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
...
+      activity.activity
+    ) || title
+  );
+}
+
+function normalizeIconName(value: string): string {
+  return value.toLowerCase().replace(/[^a-z0-9]+/g, '');
+}
+
+function getMacAppBundlePath(processPath: string | null): string | null {
```
casaout: can we move such methods that we might want to use elsewhere (incl. getting icons etc) to a helper js, to reduce the size of this already huge file?

casaout (changes requested):

src/electron/electron/main/__tests__/RetrospectionService.test.ts
grigor-dochev: It extends the Electron mock because R`etrospectionService.ts` now imports `app.getFileIcon` and `nativeImage.createFromPath` for hover icon resolution. The tests still only cover title-cleaning helpers; the mock just prevents Jest from failing when importing the service outside Electron.

src/electron/src/components/StackedBarChart.vue
```diff
@@ -39,102 +45,224 @@ const props = defineProps({
...
   activeTaskTotalTimeSpentPerActivityGroup.forEach((item) => {
-    item.timeInPercentage = (item.totalTime / totalTimeSpent) * 100
-  })
-  return activeTaskTotalTimeSpentPerActivityGroup
+    item.timeInPercentage = (item.totalTime / totalTimeSpent) * 100;
+  });
+  return activeTaskTotalTimeSpentPerActivityGroup;
+}
+
+function escapeHtml(value: string): string {
```
casaout: should this function be moved into a helper maybe?

src/electron/src/components/StackedBarChart.vue
```diff
@@ -39,102 +45,224 @@ const props = defineProps({
...
+
+  return `
+    <ol style="margin: 8px 0 0; padding: 8px 0 0; border-top: 1px solid rgba(148, 163, 184, 0.35); list-style: none; text-align: left;">
+      ${detailRows}
+      ${hiddenRow}
+    </ol>
+  `;
+}
+
+function clamp(value: number, min: number, max: number): number {
```
casaout: move into helpers?

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -637,6 +689,174 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
   );
 }
 
+// Short display label shown inside the tooltip row; keeps paths/domains compact.
```
casaout: please use the proper, unified formatting to create function comments

/**

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -637,6 +689,174 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
...
+function getTimelineHoverTitle(activity: WindowActivityEntity): string | null {
```
casaout: for getTimelineHoverTitle and getTimelineHoverTooltipTitle, it seems the key difference is long/short versions of the title; i think then the funciton names should be refined

src/electron/electron/main/services/RetrospectionService.ts
```diff
@@ -637,6 +689,174 @@ function isRelevantTopItem(session: ActivitySessions): boolean {
...
+      end,
+      activeMinutesSet,
+      date
+    );
+  }
+
+  return spans;
+}
+
+function getOverlapDurationMs(
```
casaout: move to helpers?

casaout (approved):

casaout (approved):

PR #556, opened 2026-07-06
[#554] Experience Sampling: Allow Multiple Questions

### LINKED ISSUE #554: Experience Sampling: Allow multiple questions
So far, when adding several questions in the `experienceSamplingTracker`, one was randomly picked. 

However, there are usecases when a researcher wants to ask all of them.

Thus, we'd like to introduce a new property: `showAllQuestionsTogether` (Default: false)

Behavior when `true`:
- it shows the experiencesampling pop-up with all questions stacked after each other, in the order they are defined in study.config.ts. (please also automatically update the window's height accordingly)
- To know when to actually close the window, we'll show the "save" button (same one as used for open text responses)

Behavior when `false`
behave as-is (i.e. randomly show one of the questions).

grigor-dochev:
# Experience Sampling: Allow Multiple Questions (Closes #554)

## Description

Adds a `showAllQuestionsTogether` option to the ExperienceSamplingTracker. When enabled, the experience sampling popup shows all configured questions in one stacked prompt and requires all questions to be answered before saving.

### Changes Made

- Added `showAllQuestionsTogether` to the experience sampling tracker configuration
- Keeps the default behavior unchanged by randomly showing one configured question
- Shows all configured questions in order when `showAllQuestionsTogether` is enabled
- Requires every displayed question to be answered before Save is enabled
- Saves all responses from the popup together through a batch persistence path
- Records skipped rows for all displayed questions when the popup is skipped
- Automatically resizes the experience sampling window for stacked questions, capped to the screen height
- Documents the new configuration option for researchers

## Screenshots

| All questions stacked |
|:-:|
| Experience sampling popup with all questions stacked |

Closes #554


documentation/RESEARCH.md
```diff
@@ -78,7 +78,8 @@ You can find more information on electron-builder here: https://www.electron.bui
...
+| `questions`             | An array of questions that the participant should self-reflect on. You can define one or multiple questions within the array. By default, if the array consists of multiple questions, one question will be randomly selected for each prompt.                                                                                                                                                                                                                                                          | ✅               |                           |
+| `showAllQuestionsTogether` | When set to `true`, the experience sampling popup shows all configured questions stacked in their defined order and requires all answers before the participant can save. When set to `false`, one configured question is randomly selected for each prompt.                                                                                                                                                                                                                                           |                 | `false`                   |
```
casaout: `and requires all answers before the participant can save`

That's actually a great point to discuss: i was thinking of also adding a requireAllAnswers flag (like for the end/start-of-day pop-up). However, I don't think we should require all answers for experience sampling, as chances are high that people will just click through.

Thus, I suggest to enable the Save-Button once at least one response was given

src/electron/src/views/ExperienceSamplingView.vue
```diff
@@ -2,205 +2,405 @@
...
-  selectedQuestion.answerType === 'LikertScale'
-    ? Array.from({ length: selectedQuestion.scale }, (_, i) => i + 1)
-    : [];
-const choiceOptions =
-  selectedQuestion.answerType === 'SingleChoice' || selectedQuestion.answerType === 'MultiChoice'
-    ? selectedQuestion.responseOptions
-    : [];
-const useChoiceDropdown = computed(() => choiceOptions.length >= 10);
-const choiceSelectSize = computed(() => Math.min(Math.max(choiceOptions.length, 6), 10));
+const displayedQuestions = computed<ExperienceSamplingQuestion[]>(() => {
```
casaout: displayedQuestions depends only on showAllQuestionsTogether and randomQuestionNr, both of which are constants evaluated once at setup — so this computed never recomputes. It (and hasQuestions on line 26) could just be plain constants. Not a bug, but the reactivity here is misleading: it signals the value can change when it can't. Same underlying smell as the per-render getters below (getScale, getChoiceOptions, etc.) — these run on every render for every question instead of being memoized, which is what pushed the logic out of computeds and into the template.

(From Claude)

src/electron/src/views/ExperienceSamplingView.vue
```diff
@@ -2,205 +2,405 @@
...
+    const questions = displayedQuestions.value;
+    if (showAllQuestionsTogether) {
+      const responseInputs = questions.map((question, index) =>
+        buildResponseInput(question, index, true)
+      );
+      await Promise.all([
+        typedIpcRenderer.invoke('createExperienceSamples', promptedAt, responseInputs, trigger),
+        new Promise((resolve) => setTimeout(resolve, 150))
+      ]);
+    } else if (questions[0]) {
```
casaout: The single-question branch re-implements the same createExperienceSample IPC call that already exists in createExperienceSample() (just with skipped=true). Consider routing both skip and answer through one helper, or building the single-question skip via buildResponseInput(question, 0, true) + createExperienceSamples so there's a single persistence path. Would remove ~15 lines of duplication.

(from Claude)

src/electron/src/views/ExperienceSamplingView.vue
```diff
@@ -213,141 +413,172 @@ async function skipExperienceSample() {
...
+                    >
+                      {{ option }}
+                    </option>
+                  </select>
+                </div>
               </div>
             </div>
           </div>
         </div>
+        <div v-else class="empty-state">No self-reflection questions configured.</div>
```
casaout: great to capture that edge case. i, however think that in case no self-reflection question is configured, we should silently fail the pop-up; i.e. not open it and add a log-statement

casaout (changes requested):
great job, thank you! - just a few minor comments

PR #558, opened 2026-07-09
[#551] Retrospection: Show Self-Report Legend for Single Question

### LINKED ISSUE #551: Bug: Self-Report Visualization misses Legend
When the self-report data contains only one question, no legend is shown:

Image

(related to #511 )

grigor-dochev:
# Retrospection: Show Self-Report Legend for Single Question (Closes #551)

## Description

Fixes the self-report timeline visualization so the legend is shown even when the day contains responses for only one self-report question. Previously, the legend was hidden unless there were multiple question series, which made the single-question chart harder to interpret.

### Changes Made

- Shows the self-report legend whenever at least one self-report series exists
- Removes unreachable tooltip suffix code in the same component

## Screenshots

| Single-question self-report legend |
|:-:|
| Screenshot 2026-07-09 at 23 38 54 |

Closes #551

### DIFF
--- a/src/electron/src/components/SelfReportTimelineChart.vue
+++ b/src/electron/src/components/SelfReportTimelineChart.vue
@@ -5,7 +5,7 @@
       class="self-report-tooltip rounded border border-gray-200 bg-white p-2 text-gray-700 opacity-0 shadow-lg transition-opacity duration-300 ease-in-out dark:border-transparent dark:bg-neutral-800 dark:text-neutral-300 dark:shadow-neutral-800/80"
     ></div>
     <svg ref="chart" :width="svgWidth" :height="svgHeight"></svg>
-    <div v-if="series.length > 1" class="self-report-legend">
+    <div v-if="series.length > 0" class="self-report-legend">
       <button
         v-for="item in series"
         :key="item.question"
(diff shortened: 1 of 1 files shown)


src/electron/src/components/SelfReportTimelineChart.vue
```diff
@@ -5,7 +5,7 @@
       class="self-report-tooltip rounded border border-gray-200 bg-white p-2 text-gray-700 opacity-0 shadow-lg transition-opacity duration-300 ease-in-out dark:border-transparent dark:bg-neutral-800 dark:text-neutral-300 dark:shadow-neutral-800/80"
     ></div>
     <svg ref="chart" :width="svgWidth" :height="svgHeight"></svg>
-    <div v-if="series.length > 1" class="self-report-legend">
+    <div v-if="series.length > 0" class="self-report-legend">
```
casaout: isn't this condition a tautology? i.e. isn't a length always bigger than 0?

casaout (changes requested):

src/electron/src/components/SelfReportTimelineChart.vue
grigor-dochev: I think you're right, the actual bug was the old `series.length > 1` guard, which hid the legend for a single question. So removing that condition means the `v-for` controls the rendered legend items and now they should always show.

PR #561, opened 2026-07-14
 [#555] Fix retrospection visualization data

### LINKED ISSUE #555: Retrospection: Evaluate & implement Improvements suggested by Fable
I ran Fable to check on the recently added features of the Retrospection and it made several suggestions, some (e.g. that all data queries/visualizations should consider the 4am-shift sound very reasonable).

Please evaluate if they make sense, *before* implementing them. Looking forward to your feedback on which ones made sense and which one not.

# Retrospection data visualizations — review findings

Scope: `RetrospectionView.vue`, `StackedBarChart.vue`, `SelfReportTimelineChart.vue`, `utils/retrospection/utils.ts`, `utils/retrospection/types.ts` (+ `RetrospectionService.ts` / `IpcHandler.ts` / `ExperienceSamplingService.ts` traced for data shape). Ranked by severity; no crash-level bugs found.

### 1. Self-reports are fetched for the wrong day: no 04:00 workday shift + UTC date off-by-one (wrong data)

`ExperienceSamplingService.ts:61-74`, triggered from `RetrospectionView.vue:397-400` (`parseDateInputValue`)

`getExperienceSamplingDtosForDay` is the **only** retrospection data source that does not resolve the day via `getRetrospectionWorkdayRange` (every query in `RetrospectionService.ts` does). It matches the plain local calendar date, and computes that date incorrectly:

- **(a) Calendar day instead of 04:00→04:00 workday:** a self-report prompted at 00:30 is attributed to the new calendar day, while the surrounding window activity belongs to the previous workday. Affects every user working past midnight: the report is missing from the previous day's retrospection and lands on the next day's chart, where it falls before the activity-derived x-domain and is clipped (see finding 3).
- **(b) `toISOString()` off-by-one (`ExperienceSamplingService.ts:65`):** the day string is derived from the UTC date. When a date is picked via the date input (local midnight), any UTC+ timezone (e.g. Zurich) yields the *previous* date — the chart then shows the previous day's self-reports next to the selected day's activities.

**Fix (resolves both):** query `promptedAt` within the workday range, same pattern as `getWindowActivities` (`RetrospectionService.ts:222-235`). `getExperienceSamplingDtosForDay` has exactly one caller (retrospection IPC), so the semantics change is safe:

```ts
const range = getRetrospectionWorkdayRange(d);
...
  .where("datetime(experienceSampling.promptedAt, 'localtime') >= :start", {
    start: formatSqliteLocalDateTime(range.start)
  })
  .andWhere("datetime(experienceSampling.promptedAt, 'localtime') < :end", {
    end: formatSqliteLocalDateTime(range.end)
  })
```

### 2. `InformalMeeting` time silently missing from timeline, legend, and donut (wrong data)

`RetrospectionService.ts:1024-1040` vs. `utils.ts:38`

- **Trigger:** any day containing `InformalMeeting` window activity.
- **Wrong behavior:** the renderer groups `Meeting: [PlannedMeeting, InformalMeeting]`, but the `excludeUnspecificActivities` whitelist includes `'PlannedMeeting'` and omits `'InformalMeeting'`, so those sessions never reach the renderer — the time vanishes from the stacked bar, legend totals, breakdown donut, and "Spanning work" bounds. The filter is documented as excluding only *unspecific* activities ("Other"), so this is an oversight.
- **Fix:** add `'InformalMeeting'` to the whitelist.

### 3. Self-report points outside the activity-derived x-domain are silently clipped (missing data)

`RetrospectionView.vue:124-130` (`selfReportTimelineStartDate/EndDate`), rendered by `SelfReportTimelineChart.vue:281, 333-334`

- **Trigger:** when activity data exists, the self-report chart's domain is the *activity* window only. Reports can fall outside it: (a) the workday/calendar mismatch of finding 1 — near-guaranteed until that is fixed; (b) even afterwards, activity extents come from the *filtered* specific activities, so reports prompted during `Other`/`Unknown`/unclassified app time at the day's edges don't extend the domain.
- **Wrong behavior:** `x(point.promptedAt)` lands at cx < 0 or cx > width and the SVG clips it — points and line segments silently disappear; in the worst case the "Self-reports over time" header renders with an empty chart although `hasLikertSelfReports` is true.
- **Fix:** pass the already-computed union `timelineStartDate`/`timelineEndDate` (`RetrospectionView.vue:108-122`, currently only a fallback) to the self-report chart — and to `StackedBarChart` too if the two x-axes should stay aligned.

### 4. Garbled tooltip text "…on a scale from 1-7Scale: …" plus unreachable `return` (visual, every hover)

`SelfReportTimelineChart.vue:391-397`

- **Trigger:** hovering any self-report point whose question has `responseOptions` labels — every point with the current study config (both active Likert questions define labels).
- **Wrong behavior:** `getTooltipScaleSuffix` returns `` `Scale: ${labels.join(' / ')}` ``, concatenated with no separator: "Rating: 5 on a scale from 1-7Scale: not at all productive / moderately productive / very productive". The second `return` — `` ` (${labels.join(' / ')})` ``, clearly the intended format — is unreachable dead code.
- **Fix:** delete the first `return` (line 395) so the parenthesized variant is used.

### 5. A failed activities IPC call leaves the view stuck in loading limbo (error path)

`RetrospectionView.vue:230-240` (`loadData`), `262-268` (`loadWindowActivities`)

- **Trigger:** `retrospectionGetActivities` rejects (only retrospection handler without try/catch in `IpcHandler.ts:337`, e.g. DB error); `loadWindowActivities` is the only loader in the view without its own try/catch.
- **Wrong behavior:** `loadData` throws before `isLoading.value = false`, so `isLoading` stays true forever: both charts stay hidden, the "No data" empty state can never appear, and the tiles mix the new day's top apps/websites with the previous day's donut data.
- **Fix:** wrap the `loadData` body in `try { … } finally { isLoading.value = false; }` (and/or add the missing try/catch to `loadWindowActivities`).

### 6. Latent: `StackedBarChart` data watcher rebuilds with stale domain and broken join

`StackedBarChart.vue:64-82, 96-98, 452-474`

Currently unreachable — every reload toggles `isLoading`, which remounts the chart via `v-if`, so fresh props always flow through `onMounted`. But if the watcher ever fires (e.g. the `v-if` gate is later removed), `rebuildChartWithAnimation` uses `chartStartDate`/`chartEndDate` frozen at mount (old day's domain), different margins than `buildChart` (20/20 vs 0/0 → 720 px vs 760 px scale), and a data join with no `.enter()` — bars for the new day are mispositioned or never drawn.

**Fix:** recompute the clamped domain from current props inside the watcher, or delete the watcher/animation path since remounting is the actual update mechanism.

---

**Convention (documented in code):** all retrospection data queries **and** visualizations must resolve "a day" uniformly through the 04:00→04:00 workday range — noted in the `getRetrospectionWorkdayRange` docstring (`RetrospectionService.ts:150-160`) and above `loadData()` (`RetrospectionView.vue:227-229`). Finding 1 is the only current violator.

---

## Focus areas checked, no real bug found

1. **Number/time math** — no reachable division-by-zero or NaN: the `MAX_SAFE_INTEGER`/`0` reduce sentinels are gated by `hasActivityData`/`hasLikertSelfReports` before use, `getTopItemWidth` clamps its denominator to ≥ 1, and the legend's `timeInPercentage` (NaN on empty data) is computed but never displayed. Only nit: with a 2-point scale, `getScaleAxisLabels` (`SelfReportTimelineChart.vue:188-195`) draws the mid label "1" at value 0, exactly overlapping the low label — cosmetic, and impossible with the current config (scales 5 and 7).
2. **Donut denominator vs. numerator** — consistent: the service already filters `Other`/`Unknown`/`Idle` out of `retrospectionGetActivities`, so the view's `EXCLUDED_ACTIVITY_BREAKDOWN_GROUPS` check (`RetrospectionView.vue:37`) is a redundant no-op; segment percentages are fractions of the same total the segments come from (can't exceed 360°), and the neutral tail correctly represents the top-6/90% cutoff remainder.
3. **Dead/unreachable code & string building** — one instance found, reported as finding 4 (double `return` in `getTooltipScaleSuffix`).
4. **D3 lifecycle** — `matchMedia` listeners are removed in `onUnmounted` in both charts, tooltips are single instances, and the remount-per-reload pattern (`v-if="!isLoading …"`) prevents duplicate bars/legends/axes; the latent watcher path is finding 6. Micro-nit: a theme change rebuilds `StackedBarChart` with all bars at full opacity but doesn't reset `chartSelectedLegendItem` (`StackedBarChart.vue:58-62` vs. `452-453`), so the first click on the previously focused legend item is a visual no-op.
5. **Grouping/color/label lookups** — every group `getActivityGroupFromActivityName` can return has entries in both `TW_CLASS_ACTIVITY_MAPPINGS` and `ACTIVITY_LABELS`, with an `'Other'` fallback, so no undefined fill or "undefined" text is reachable renderer-side; the only gap is service-side (finding 2, `InformalMeeting` never delivered).
6. **XSS / HTML injection** — clean. Every OS/user-derived string in the `StackedBarChart` innerHTML tooltip (`title`, `tooltipTitle`, `appName`, `iconDataUrl`, the fallback initial) passes through `escapeHtml`, which correctly escapes `&` first plus both quote styles; remaining interpolations are internal constants (colors, labels, times). `SelfReportTimelineChart` builds its tooltip via `textContent`/`replaceChildren`; everything else is Vue-escaped interpolation.
7. **Empty/edge data** — solid: the service always sets both `from` and `to` on sessions (the last open row gets a 1-minute fallback), so the `!activity.end → Date.now()` branch (`RetrospectionView.vue:59-61`) is dead defensive code with current data; all-skipped/invalid reports fall through to the "No self-reports" card; non-JSON `responseOptions` is caught in `parseLikertLabels`; a day with only self-reports (or only activities) renders via the `has*` gates.
8. **Timezone/date correctness** — the real issues are consolidated into findings 1 (calendar-day + UTC off-by-one in the self-report query) and 3 (domain clipping). Otherwise fine: `parseDateInputValue`/`formatDateInputValue` are a symmetric local-time pair, the `StackedBarChart` `onMounted` clamp behaves for partial "today" and midnight-spanning workdays, and `scaleTime` handles DST days. Design note, not a bug: `isToday` is calendar-based while data uses the 04:00 cutoff, so opening the retrospection between 00:00 and 04:00 shows "No data for this day" for "Today" — that night's work is intentionally under "Yesterday".

Note: two items above are borderline actionable rather than pure non-issues — the scale-2 label overlap (item 1) and the legend-selection reset after theme change (item 4). Both are cosmetic edge cases; promote them to findings if they should be tracked as tasks.

grigor-dochev:
# Retrospection visualization data fixes (Closes #555)

## Description

Fixes workday-boundary inconsistencies, missing informal meetings, timeline clipping, and loading failure handling in retrospection.

### Changes Made

- Resolve self-reports using the shared 04:00–04:00 retrospection workday range.
- Add shared workday and timeline-bound helpers.
- Include `InformalMeeting` in the Meeting timeline, legend, and activity breakdown.
- Use shared activity/self-report timeline bounds so valid self-report points are not clipped.
- Ensure failed activity loading clears the loading state and presents an empty state safely.
- Add workday and self-report loading regression coverage.

## Testing

- `vue-tsc --noEmit` passed.
- Production build passed.

### DIFF
--- a/src/electron/shared/retrospection/Workday.ts
+++ b/src/electron/shared/retrospection/Workday.ts
@@ -1,0 +1,89 @@
+export const RETROSPECTION_WORKDAY_CUTOFF_HOUR = 4;
+
+export interface RetrospectionWorkdayRange {
+  start: Date;
+  end: Date;
+}
+
+export interface RetrospectionTimelineBounds {
+  start: number;
+  end: number;
+}
+
+function toLocalDate(date: Date | string): Date {
+  if (date instanceof Date) return date;
+  const parts = /^(\d{4})-(\d{2})-(\d{2})$/.exec(date);
+  return parts
+    ? new Date(Number(parts[1]), Number(parts[2]) - 1, Number(parts[3]))
+    : new Date(date);
+}
+
+/** Returns the selected local calendar day's 04:00-to-04:00 workday. */
+export function getRetrospectionWorkdayRange(date: Date | string): RetrospectionWorkdayRange {
+  const d = toLocalDate(date);
+  const start = new Date(
+    d.getFullYear(),
+    d.getMonth(),
+    d.getDate(),
+    RETROSPECTION_WORKDAY_CUTOFF_HOUR,
+    0,
+    0,
+    0
+  );
+  const end = new Date(start);
+  end.setDate(end.getDate() + 1);
+  return { start, end };
+}
+
+export function getWorkdayMinuteIndex(date: Date, workdayStart: Date): number {
+  return Math.floor((date.getTime() - workdayStart.getTime()) / 60000);
+}
+
+export function getDateFromWorkdayMinuteIndex(minuteIndex: number, workdayStart: Date): Date {
+  return new Date(workdayStart.getTime() + minuteIndex * 60000);
+}
+
+/** Formats a local timestamp for SQLite `datetime(..., 'localtime')` comparison. */
+export function formatSqliteLocalDateTime(date: Date): string {
+  const parts = [
+    date.getFullYear(),
+    `${date.getMonth() + 1}`.padStart(2, '0'),
+    `${date.getDate()}`.padStart(2, '0')
+  ];
+  const time = [date.getHours(), date.getMinutes(), date.getSeconds()]
+    .map((value) => `${value}`.padStart(2, '0'))
+    .join(':');
+  return `${parts.join('-')} ${time}`;
+}
+
+function floorToLocalHour(timestamp: number): number {
+  const date = new Date(timestamp);
+  date.setMinutes(0, 0, 0);
+  return date.getTime();
+}
+
+/** Computes one hour-aligned x-axis domain for every retrospection timeline. */
+export function getRetrospectionTimelineBounds(
+  selectedDay: Date | string,
+  timestamps: Array<Date | number | string | null | undefined>
+): RetrospectionTimelineBounds | null {
+  const values = timestamps
+    .map((timestamp) =>
+      timestamp instanceof Date
+        ? timestamp.getTime()
+        : typeof timestamp === 'string'
+          ? new Date(timestamp).getTime()
+          : timestamp
+    )
+    .filter((timestamp): timestamp is number => typeof timestamp === 'number' && Number.isFinite(timestamp));
+  const { start: workdayStartDate, end: workdayEndDate } = getRetrospectionWorkdayRange(selectedDay);
+  const workdayStart = workdayStartDate.getTime();
+  const workdayEnd = workdayEndDate.getTime();
+  const inRange = values.filter((timestamp) => timestamp >= workdayStart && timestamp < workdayEnd);
+  if (!inRange.length) return null;
+
+  const start = Math.max(floorToLocalHour(Math.min(...inRange)), workdayStart);
+  let end = Math.min(floorToLocalHour(Math.max(...inRange)) + 60 * 60 * 1000, workdayEnd);
+  if (end <= start) end = Math.min(start + 60 * 60 * 1000, workdayEnd);
+  return { start, end };
+}
--- a/src/electron/electron/main/__tests__/ExperienceSamplingService.test.ts
+++ b/src/electron/electron/main/__tests__/ExperienceSamplingService.test.ts
@@ -1,0 +1,61 @@
+import { beforeEach, expect, jest, test } from '@jest/globals';
+
+const where = jest.fn();
+const andWhere = jest.fn();
+const orderBy = jest.fn();
+const getMany = jest.fn();
+const createQueryBuilder = jest.fn();
+
+jest.unstable_mockModule('electron', () => ({
+  default: {
+    app: {
+      isPackaged: false
+    }
+  }
+}));
+
+jest.unstable_mockModule('../entities/ExperienceSamplingResponseEntity', () => ({
+  ExperienceSamplingResponseEntity: {
+    createQueryBuilder
+  }
+}));
+
+const { ExperienceSamplingService } = await import('../services/ExperienceSamplingService');
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  getMany.mockResolvedValue([] as never);
+  orderBy.mockReturnValue({ getMany });
+  andWhere.mockReturnValue({ orderBy });
+  where.mockReturnValue({ andWhere });
+  createQueryBuilder.mockReturnValue({ where });
+});
+
+test('self-reports are queried using the selected 04:00-to-04:00 local workday', async () => {
+  const service = new ExperienceSamplingService();
+
+  await service.getExperienceSamplingDtosForDay(new Date(2026, 5, 29, 0, 0));
+
+  expect(createQueryBuilder).toHaveBeenCalledWith('experienceSampling');
+  expect(where).toHaveBeenCalledWith(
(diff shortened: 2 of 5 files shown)


src/electron/shared/retrospection/Workday.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function toLocalDate(date: Date | string): Date {
+  if (date instanceof Date) return date;
+  const parts = /^(\d{4})-(\d{2})-(\d{2})$/.exec(date);
+  return parts
+    ? new Date(Number(parts[1]), Number(parts[2]) - 1, Number(parts[3]))
+    : new Date(date);
+}
+
+/** Returns the selected local calendar day's 04:00-to-04:00 workday. */
```
casaout, line 21: just to clarify; with the changes does this mean that for all statistics (timelines, time spent in apps, websites etc), we're considering a workday from 04:00 to 03:59 the next day?

casaout (approved):

src/electron/shared/retrospection/Workday.ts
grigor-dochev: Yes exactly, now all retrospection data is interpreted using the selected day’s 04:00 (inclusive) to next-day-03:59 workday range. This means we cover the: 
- Activity timeline, active hours
- Longest active period
- Top apps
- Websites
- Window titles
- Activity breakdown
- Self-reports

Issue #41, opened 2016-02-18
Update Active View with the average of the last 5 work days


casaout:

casaout:
Closing as reported for legacy version of PersonalAnalytics.

PR #562, opened 2026-07-14
[#527] Extract retrospection figures into dedicated modules

### LINKED ISSUE #527: Refactor & Refine Retrospection

grigor-dochev:
# Refactor RetrospectionService (Closes #527)

## Description

Refactors the growing retrospection service into focused figure modules while preserving the existing service API for callers and tests.

### Changes Made

- Extract activity timeline, active-hours, top-app, top-website, and top-window-title figure logic into dedicated modules.
- Add a shared retrospection snapshot that fetches workday activity and user-input data once per dashboard load.
- Move window-title cleanup, activity-session construction, and timeline-detail logic into focused helpers.
- Keep `RetrospectionService` as a thin orchestration and compatibility façade.
- Isolate figure failures so one unavailable figure does not prevent the remaining dashboard data from loading.
- Add snapshot and figure-focused tests.

## Testing

- `vue-tsc --noEmit` passed.
- Production build passed.

Closes #527

### DIFF
--- a/src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
+++ b/src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
@@ -40,7 +40,11 @@ async function addWindowActivityDetailSpan(
   }
 
   const tooltipTitle = getLongTimelineHoverTitle(activity, title);
-  const iconDataUrl = await getProcessIconDataUrl(activity.processPath, activity.processName);
+  const iconDataUrl = await getProcessIconDataUrl(
+    activity.processPath,
+    activity.processName,
+    activity.processId
+  );
   getActiveMinuteSpans(from, to, workdayData.activeMinutes, workdayData.workdayStart).forEach(
     (span) => {
       spans.push({
--- a/src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
+++ b/src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
@@ -1,3 +1,7 @@
+/**
+ * Builds the per-session detail rows shown when hovering an activity-timeline bar. It is specific
+ * to the activity timeline; the other figures do not use these hover details.
+ */
 import type { WindowActivityEntity } from '../../entities/WindowActivityEntity';
 import type {
   ActivitySessions,
@@ -6,7 +10,7 @@ import type {
 import { getWorkdayMinuteIndex } from '../../../../shared/retrospection/Workday';
 import { getProcessIconDataUrl } from '../utils/AppIconHelper';
 import { getOverlapDurationMs } from '../utils/helpers';
-import type { RetrospectionSnapshot } from './RetrospectionSnapshot';
+import type { RetrospectionWorkdayData } from './RetrospectionWorkdayData';
 import { getActiveMinuteSpans } from './WindowActivitySessions';
 import { getLongTimelineHoverTitle, getShortTimelineHoverTitle } from './WindowTitle';
 
@@ -28,7 +32,7 @@ async function addWindowActivityDetailSpan(
   activity: WindowActivityEntity,
   from: Date,
   to: Date,
-  snapshot: RetrospectionSnapshot
+  workdayData: RetrospectionWorkdayData
 ): Promise<void> {
   const title = getShortTimelineHoverTitle(activity);
   if (!title || to.getTime() <= from.getTime()) {
@@ -37,29 +41,31 @@ async function addWindowActivityDetailSpan(
 
   const tooltipTitle = getLongTimelineHoverTitle(activity, title);
   const iconDataUrl = await getProcessIconDataUrl(activity.processPath, activity.processName);
-  getActiveMinuteSpans(from, to, snapshot.activeMinutes, snapshot.workdayStart).forEach((span) => {
-    spans.push({
-      from: span.from,
-      to: span.to,
-      activity: activity.activity,
-      title,
-      appName: activity.processName,
-      tooltipTitle,
-      iconDataUrl
-    });
-  });
+  getActiveMinuteSpans(from, to, workdayData.activeMinutes, workdayData.workdayStart).forEach(
+    (span) => {
+      spans.push({
+        from: span.from,
+        to: span.to,
+        activity: activity.activity,
+        title,
+        appName: activity.processName,
+        tooltipTitle,
+        iconDataUrl
+      });
+    }
+  );
 }
 
 async function getWindowActivityDetailSpans(
-  snapshot: RetrospectionSnapshot
+  workdayData: RetrospectionWorkdayData
 ): Promise<WindowActivityDetailSpan[]> {
   const spans: WindowActivityDetailSpan[] = [];
   let lastWindowActivity: WindowActivityEntity | undefined;
 
-  for (const activity of snapshot.windowActivities) {
+  for (const activity of workdayData.windowActivities) {
     if (
-      !snapshot.activeMinutes.has(
-        getWorkdayMinuteIndex(new Date(activity.ts), snapshot.workdayStart)
+      !workdayData.activeMinutes.has(
+        getWorkdayMinuteIndex(new Date(activity.ts), workdayData.workdayStart)
       )
     ) {
       continue;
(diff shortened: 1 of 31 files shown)


src/electron/electron/main/services/retrospection/figures/ActivityTimelineFigure.ts
```diff
@@ -0,0 +1,38 @@
+import { Activity, type ActivitySessions } from '../../../../../src/utils/retrospection/types';
+import type { RetrospectionSnapshot } from '../RetrospectionSnapshot';
+import { addTimelineHoverDetails } from '../TimelineDetails';
+import { getWindowActivitySessionsByType } from '../WindowActivitySessions';
+
+const SPECIFIC_ACTIVITY_TYPES = new Set<Activity>([
```
casaout, line 9: (minor) i wonder if we need to redelcare them or if we could just use the existing type?

src/electron/electron/main/services/retrospection/RetrospectionWorkdayData.ts
```diff
@@ -0,0 +1,83 @@
+import { UserInputEntity } from '../../entities/UserInputEntity';
```
casaout, line 5: can you maybe document this file/class a bit, to clarify its responsibiltiy?

also, I recommend not calling it `snapshot`, as we'll be using that notion for implementing TaskSnap

src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
```diff
@@ -0,0 +1,150 @@
...
+import { getProcessIconDataUrl } from '../utils/AppIconHelper';
+import { getOverlapDurationMs } from '../utils/helpers';
+import type { RetrospectionSnapshot } from './RetrospectionSnapshot';
+import { getActiveMinuteSpans } from './WindowActivitySessions';
+import { getLongTimelineHoverTitle, getShortTimelineHoverTitle } from './WindowTitle';
+
+const TIMELINE_HOVER_DETAIL_LIMIT = 4;
+const MIN_TIMELINE_HOVER_DETAIL_DURATION_MS = 60_000;
+
+interface WindowActivityDetailSpan {
```
casaout, line 20: is TimelineDetails basically the overlay on Hover? 
is it used just for the activityTimeline or for other too?

i think documenting that, and if it's only for the activitytimeline, also renaming it would help understanding

src/electron/electron/main/services/retrospection/TopItemIcons.ts
```diff
@@ -0,0 +1,25 @@
+import type { ActivitySessions } from '../../../../src/utils/retrospection/types';
```
casaout, line 5: what does this file/class do?

src/electron/electron/main/services/retrospection/figures/ActivityTimelineFigure.ts
grigor-dochev: The existing Activity type also includes unspecific values such as Other, Idle, and Unknown, which should be excluded here. I kept the explicit subset and added a comment explaining this.

src/electron/electron/main/services/retrospection/RetrospectionWorkdayData.ts
grigor-dochev: It loads the shared tracker data for one retrospection workday. 

I renamed it to `RetrospectionWorkdayData` to avoid conflicting with TaskSnap and documented this.

src/electron/electron/main/services/retrospection/ActivityTimelineDetails.ts
grigor-dochev: Yes, it provides the hover details for the activity timeline only. I renamed it to `ActivityTimelineDetails`.

src/electron/electron/main/services/retrospection/TopItemIcons.ts
grigor-dochev: It attaches application icons to the Top Apps, Top Websites, and Top Window Titles data.

src/electron/electron/main/services/retrospection/WindowActivitySessions.ts
```diff
@@ -0,0 +1,163 @@
+/**
```
casaout, line 1: this class is amazing; the "inactive minutes", is this with a minimum duration of e.g. 2/5 minutes and if yes, where is it defined?

(I realize that's actually not part of the port; but if it's not implemented, I'll need to create an issue on that; but maybe that's already implemented)

thanks!

src/electron/electron/main/services/retrospection/WindowTitle.ts
```diff
@@ -0,0 +1,306 @@
+/**
+ * Normalizes window titles and URLs into readable labels used by timeline hover details and the
```
casaout, line 2: will this file be simplified or even ommitted through the recent merge of https://github.com/HASEL-UZH/PA.WindowsActivityTracker/pull/41?

src/electron/electron/main/services/retrospection/WindowActivitySessions.ts
grigor-dochev: Yes, it is already implemented, but there is currently no 2/5-minute minimum period. So, a minute is active only when the User Input Monitor recorded a click, key press, scroll, or mouse movement in that minute. Every inactive minute splits the reconstructed window-activity session.

I think if we want to implement a tolerance, that should be a new issue.

src/electron/electron/main/services/retrospection/WindowTitle.ts
grigor-dochev: It is partly simplified, but not completely omitted: https://github.com/HASEL-UZH/PA.WindowsActivityTracker/pull/41(#41) moves collection-time artifact extraction into `WindowsActivityTracker` and stores `artifactName`. The retrospection now uses that value first. 

`WindowTitle.ts` remains responsible for renderer-specific browser/URL title normalisation (it also has a fallback for older rows that have no `artifactName`).

src/electron/electron/main/services/retrospection/WindowActivitySessions.ts
casaout: https://github.com/HASEL-UZH/PA.WindowsActivityTracker/issues/1

casaout (approved):

PR #563, opened 2026-07-14
[#527] Handle unavailable retrospection data sources

### LINKED ISSUE #527: Refactor & Refine Retrospection

grigor-dochev:
# Handle unavailable retrospection data sources (Closes #527)

## Description

Makes retrospection resilient when one or more required trackers are disabled, unavailable, or return incomplete data.

### Changes Made

- Replace multiple independent retrospection IPC calls with one dashboard request.
- Return available dashboard sections even when another section fails to load.
- Show clear empty-state and partial-loading messages for disabled trackers and failed sections.
- Reset stale data before loading a newly selected day.
- Add tracker-availability helpers and tests.
- Keep self-reports available when activity insights are unavailable.
- Reuse the top-item card treatment for Top Apps, Top Websites, and Top Window Titles.

## Screenshots

| All trackers disabled | Window activity tracking disabled | User input tracking disabled |
|:-:|:-:|:-:|
| all_trackers_disabled | windows_tracking_disabled | user_input_tracking_disabled
 
## Testing

- `vue-tsc --noEmit` passed.
- Production build passed.

Closes #527

### DIFF
--- a/src/electron/shared/study.config.ts
+++ b/src/electron/shared/study.config.ts
@@ -21,17 +21,20 @@ const studyConfig: StudyConfiguration = {
   trackers: {
     enabledWorkHours: true,
     windowActivityTracker: {
+      name: 'Window Activity Monitor',
       enabled: true,
       intervalInMs: 1000,
       trackUrls: false,
       trackWindowTitles: true
     },
     userInputTracker: {
+      name: 'User Input Monitor',
       enabled: true,
       intervalInMs: 60000,
       collectKeyDetails: true
     },
     experienceSamplingTracker: {
+      name: 'Experience Sampling',
       enabled: true,
       questions: [
         {
@@ -46,7 +49,7 @@ const studyConfig: StudyConfiguration = {
           answerType: 'LikertScale',
           scale: 5,
           responseOptions: ['not well', 'moderately well', 'very well']
-        },
+        }
         // {
         //   question: 'What is one aspect that affected your ability to focus the most in the last session?',
         //   answerType: 'TextResponse',
--- a/src/electron/src/utils/Commands.ts
+++ b/src/electron/src/utils/Commands.ts
@@ -13,7 +13,7 @@ import type {
   DailySurveySamplingType,
   ExperienceSamplingAnswerType
 } from '../../shared/StudyConfiguration';
-import type { ActiveHoursInsight, ActivitySessions, TimeActive } from './retrospection/types';
+import type { RetrospectionDashboard } from './retrospection/types';
 
 type Commands = {
   createExperienceSample: (
@@ -62,13 +62,7 @@ type Commands = {
   startAllTrackers: () => void;
   triggerPermissionCheckAccessibility: (prompt: boolean) => boolean;
   triggerPermissionCheckScreenRecording: () => boolean;
-  retrospectionGetActiveHours: (date: Date) => Promise<ActiveHoursInsight>;
-  retrospectionGetActivities: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionLoadLongestTimeActive: (date: Date) => Promise<TimeActive | undefined>;
-  retrospectionGetTopThreeMostActiveApps: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetTopThreeWebsites: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetTopThreeWindowTitles: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetSelfReports: (date: Date) => Promise<ExperienceSamplingDto[]>;
+  retrospectionGetDashboard: (date: Date) => Promise<RetrospectionDashboard>;
   openRetrospection: () => Promise<void>;
   closeRetrospectionWindow: () => void;
   createDailySurveyResponses: (
--- a/src/electron/src/utils/retrospection/availability.ts
+++ b/src/electron/src/utils/retrospection/availability.ts
@@ -1,29 +1,45 @@
 export interface RetrospectionTrackerAvailability {
   activityInsightsEnabled: boolean;
   selfReportsEnabled: boolean;
+  activityInsightMessages: string[];
+  selfReportMessages: string[];
   messages: string[];
 }
 
+interface TrackerAvailabilityConfig {
+  name: string;
+  enabled: boolean;
+}
+
 export function getRetrospectionTrackerAvailability(
-  windowActivityEnabled: boolean,
-  userInputEnabled: boolean,
-  experienceSamplingEnabled: boolean
+  windowActivityMonitor: TrackerAvailabilityConfig,
+  userInputMonitor: TrackerAvailabilityConfig,
+  experienceSampling: TrackerAvailabilityConfig
 ): RetrospectionTrackerAvailability {
-  const messages: string[] = [];
-  if (!windowActivityEnabled) {
-    messages.push('Window activity tracking is disabled, so activity insights are unavailable.');
-  } else if (!userInputEnabled) {
-    messages.push(
-      'User-input tracking is disabled, so active-time-based activity insights are unavailable.'
+  const activityInsightMessages: string[] = [];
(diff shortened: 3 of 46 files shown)


src/electron/src/utils/retrospection/availability.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function getRetrospectionTrackerAvailability(
+  windowActivityEnabled: boolean,
+  userInputEnabled: boolean,
+  experienceSamplingEnabled: boolean
+): RetrospectionTrackerAvailability {
+  const messages: string[] = [];
+  if (!windowActivityEnabled) {
+    messages.push('Window activity tracking is disabled, so activity insights are unavailable.');
```
casaout: ideally, we'd show the trackers' name, e.g. "Window Activity Monitor" for PA.WindowsActivityTracker (it's a property of the trackers)

also, we should avoid the term "tracking" and use "monitoring"

finally, instead of "unavailable", i would write something like "Windows Activity Monitor is disabled, so no activity insights can be visualized."

src/electron/src/views/RetrospectionView.vue
```diff
@@ -69,75 +97,103 @@ const latestUserComputerActivity = computed((): number => {
...
-    ...(chartDataWindowActivities.value ?? []).flatMap((point) => [point.start, point.end]),
-    ...selfReports.value.map((report) => report.promptedAt)
-  ]);
-  if (bounds) return bounds.start;
-  const candidates = [
-    hasActivityData.value ? earliestUserComputerActivity.value : null,
-    hasLikertSelfReports.value ? earliestSelfReport.value : null
-  ].filter((value): value is number => value !== null && Number.isFinite(value));
-  return Math.min(...candidates);
+const emptyStateDescription = computed((): string => {
```
casaout, line 158: how can i best trigger and test these message? (which cases do we see here?)

Note that also here, we should use the trackers' Name property, and use the word "monitoring" instead of "tracking"

I also wonder, depending on the actual errors, if it should just silently fail?

src/electron/src/utils/Commands.ts
```diff
@@ -62,13 +62,7 @@ type Commands = {
...
   triggerPermissionCheckAccessibility: (prompt: boolean) => boolean;
   triggerPermissionCheckScreenRecording: () => boolean;
-  retrospectionGetActiveHours: (date: Date) => Promise<ActiveHoursInsight>;
-  retrospectionGetActivities: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionLoadLongestTimeActive: (date: Date) => Promise<TimeActive | undefined>;
-  retrospectionGetTopThreeMostActiveApps: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetTopThreeWebsites: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetTopThreeWindowTitles: (date: Date) => Promise<ActivitySessions[]>;
-  retrospectionGetSelfReports: (date: Date) => Promise<ExperienceSamplingDto[]>;
+  retrospectionGetDashboard: (date: Date) => Promise<RetrospectionDashboard>;
```
casaout, line 65: do I understand that correctly that you're no longer loading individual datasets/figures, but the entire dashboard through one request?

how will this practice be impacted once we allow users to disable/hide certain figures? will they then still be loaded?

src/electron/src/views/RetrospectionView.vue
grigor-dochev: The messages can be tested by disabling Window Activity Monitor, User Input Monitor, Experience Sampling, or all three in `study.config.ts`. The actual loading errors remain visible so they are not mistaken for missing data.

src/electron/src/utils/Commands.ts
grigor-dochev: That's correct, yep. The dashboard is loaded through one IPC request and the raw data is shared across figures. Hidden figures would currently still be calculated; once we allow visibility to be configurable, I think the request can specify which sections should be built.

src/electron/shared/study.config.ts
```diff
@@ -21,17 +21,20 @@ const studyConfig: StudyConfiguration = {
   trackers: {
     enabledWorkHours: true,
     windowActivityTracker: {
+      name: 'Window Activity Monitor',
```
casaout, line 24: sorry, that's not what I meant.
I think we can read the property from the tracker, e.g. here: https://github.com/HASEL-UZH/PA.WindowsActivityTracker/blob/main/typescript/src/index.ts#L14

that way it'll be much easier to handle should we end up renaming them in the future

src/electron/src/views/RetrospectionView.vue
casaout: Yes, I could previously already reproduce the basic empty state messages when disabling one or more trackers.

What I mean ar the messages below (L158ff), e.g. when sections fail to load

src/electron/src/utils/retrospection/availability.ts
casaout: sorry, the updated implementation is not what I meant.

I think we can read the property from the tracker, e.g. here: https://github.com/HASEL-UZH/PA.WindowsActivityTracker/blob/main/typescript/src/index.ts#L14

that way it'll be much easier to handle should we end up renaming them in the future

src/electron/src/views/RetrospectionView.vue
grigor-dochev: Ah ok, I understood. Those messages are for actual dashboard/section loading failures, not disabled trackers. A query or figure failure is recorded in `dashboard.errors`, which then lists the affected sections and shows Retry rather than silently presenting missing data as an empty day.

casaout: How can I best "trigger"/produce these to test?

Do I understand correctly that if one fails, there is a "retry" button the use can press?

grigor-dochev: Yep exactly, the "Try again" button calls `loadData()` and reloads the dashboard for the currently selected day.

To trigger it manually, you can temporarily throw an error from a dashboard section builder, for example the Top Apps figure, then open Retrospection. That will mark the section as failed. If you then remove the temporary error and click on "Try again", it should load it normally.

casaout (approved):

PR #565, opened 2026-07-17
 [#25] Persist window activity artifact names

### LINKED ISSUE #25: Improve Detection/Extraction of Filenames
Currently, we extract file names based on window titles using simple heuristics.

Let's think about better ways to do so. Also, let's store them separately in the database for easier access and querying.

Losely related to #15

grigor-dochev:
# Persist Window Activity Artifact Names (Closes HASEL-UZH/PA.WindowsActivityTracker#25)

## Description

Persists artifact names provided by PA.WindowsActivityTracker and uses them in retrospection details and top-item visualizations.

### Changes Made

- Added nullable `artifactName` to the `window_activity` entity and DTO
- Persists and returns the captured artifact name through WindowActivityTrackerService
- Uses artifact names for timeline hover labels and top websites/window titles where available
- Includes artifact names in data-export anonymization

### DIFF
--- a/src/electron/electron/main/entities/WindowActivityEntity.ts
+++ b/src/electron/electron/main/entities/WindowActivityEntity.ts
@@ -18,6 +18,9 @@ export class WindowActivityEntity extends BaseTrackedEntity {
   @Column('text', { nullable: true })
   url: string | null;
 
+  @Column('text', { nullable: true })
+  artifactName: string | null;
+
   @Column('text', { nullable: false })
   activity: string;
 
--- a/src/electron/electron/main/services/DataExportService.ts
+++ b/src/electron/electron/main/services/DataExportService.ts
@@ -133,10 +133,11 @@ export class DataExportService {
       const items: {
         windowTitle: string;
         url: string;
+        artifactName: string | null;
         id: string;
       }[] = await WindowActivityEntity.getRepository()
         .createQueryBuilder('window_activity')
-        .select('id, windowTitle, url')
+        .select('id, windowTitle, url, artifactName')
         .getRawMany();
       for (const item of items) {
         if (windowActivityExportType === DataExportType.Obfuscate) {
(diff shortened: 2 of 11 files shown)


src/electron/electron/main/entities/WindowActivityEntity.ts
```diff
@@ -18,6 +18,9 @@ export class WindowActivityEntity extends BaseTrackedEntity {
   @Column('text', { nullable: true })
```
casaout, line 18: do I assume correctly that you were able to copy most of this from the PA-repo to the PA.WindowsActivityTracker and will now remove this code from the PA-repo to avoid duplication?

src/electron/PA.WindowsActivityTracker
casaout, line 1: would it make sense to update the PA.WindowsActivityTracker now to the commit of the approved PR?

src/electron/electron/main/entities/WindowActivityEntity.ts
grigor-dochev: Not entirely. The collection-time artifact extraction now lives in PA.WindowsActivityTracker. `WindowTitle.ts` remains as a backwards-compatible display fallback for existing rows without `artifactName`, and it also provides the short/long presentation variants used by the timeline and cards.

src/electron/PA.WindowsActivityTracker
grigor-dochev: Yes, I agree. I updated the PersonalAnalytics submodule reference to the merged PA.WindowsActivityTracker PR commit.

casaout: I might be wrong, but I think you've updated the to the wrong commit (on main, but I haven't merged the changes  there yet). 

I now just did that and updated to `6e87279fb5e4e0649cfd970731e9c5425584a2e3`

casaout (approved):

PR #566, opened 2026-07-23
[#525] Daily Survey: Preserve Original Date for Overdue Surveys

### LINKED ISSUE #525: Daily Survey: Bug when re-opening on later days
Today (5/15), I've received a daily survey as I didn't answer it on 5/13. However, it shows that it is for the day 5/14, even though I didn't use my computer.

Can you please double check how the date is fetched? (it should've shown 5/13):
(Edit: on 17/5 it correctly showed 15/5; so maybe just good if you can test the behavior as well)
Image

In the logs you see that I've last postponed it on 5/13; there are entries on 5/14 even if I didn't run my laptop, but these are automated system updates.

Image

Finally, when it's a daily survey for a previous day, it shouldn't be postponable (skip is fine)

grigor-dochev:
# Daily Survey: Preserve Original Date for Overdue Surveys (Closes #525)

## Description

Fixes daily surveys being reassigned to a newer day when PersonalAnalytics reopens after an unattended system resume or application restart.

An overdue survey now retains its original scheduled date until it is answered or skipped. Previous-day surveys show their original date and cannot be postponed.

### DIFF
--- a/documentation/RESEARCH.md
+++ b/documentation/RESEARCH.md
@@ -95,6 +95,9 @@ You can find more information on electron-builder here: https://www.electron.bui
 | `surveys[].requireAllAnswers`   | Whether all questions must be answered before the survey can be submitted. If `false`, partial submissions are allowed. Can be set independently for each survey.                                                                                                                                                                                                                         |                 | `false`       |
 | `surveys[].questions`           | An array of questions to present in the survey. All questions are shown in order. Each question uses the same format as ExperienceSamplingTracker questions and supports `answerType` values: `’LikertScale’`, `’SingleChoice’`, `’MultiChoice’`, and `’TextResponse’`.                                                                                                                    | ✅               |               |
 
+##### Overdue Daily Surveys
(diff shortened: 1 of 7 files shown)


casaout:
Few suggested improvements, committed with 39673f23540193239a1aa7e1b1761373059be01f:

**Error handling.** `processSurveys()` is called from two places that drop its promise — the node-schedule tick (node-schedule ignores the return value) and `resume()` (`TrackerService.resumeAllTrackers` discards it via the `: void` annotation on the `.map`). That was harmless while `resume()` only restarted the cron, but it now does DB reads *and writes*, so a failure disappears with no log line and half-written survey state. Realistic triggers: `loadURL` rejecting with `ERR_ABORTED` when a window is superseded or closed mid-load, the tick firing while the DataSource is torn down at shutdown, `getPrimaryDisplay()` on a machine waking undocked. Both call sites now log. No retry logic added — the every-minute tick already re-derives everything from persisted state, so a failed run self-heals. Deliberately *not* caught inside `processSurveys()` so `start()` keeps propagating startup failures.

**Cleanups.** Extracted `isActiveSurvey()` and folded the `activeSurvey` reset into `clearShownSurvey()`, replacing three copies of the same check; dropped a redundant postponed-until reset that `scheduleNextForSurvey()` already performs; threaded `now` into `isBeforeToday()` instead of re-reading the clock mid-pass; restored the doc comment that got dropped from `getPostponedUntilField()` and documented the pending-vs-next distinction, since that's the crux of the fix.

**Docs.** Added an "Overdue Daily Surveys" note to `RESEARCH.md`. The part that matters for analysis: a response's `promptedAt` can now trail the `scheduledDate` it refers to by up to 3 days, so `scheduledDate` is the column to join on when relating responses to a day.

PR #574, opened 2026-07-30
[#520] Update Node runtime to 24

### LINKED ISSUE #520: Update Node
We're currently building using Node 20, we'd like to upgrade that to 24

grigor-dochev:
# Update Node Runtime to 24 (Closes #520)

## Description

Updates the Node.js version used to install and build PersonalAnalytics from 20 to 24, and aligns Electron and the native SQLite dependencies for compatibility.

### Changes Made

- Updated the CI build environment to Node.js 24 and `actions/setup-node@v5`
- Added an `.nvmrc` file for consistent local Node.js versions
- Added the required Node.js version to `package.json`
- Updated the research documentation to require Node.js 24
- Upgraded Electron and the SQLite native dependencies
- Regenerated the package lockfile
- Updated the preload IPC bridge for Electron 29 compatibility; the renderer now receives the `invoke` method it uses rather than the native `ipcRenderer` object

## Verification

- Node.js 24.15.0: `npm ci` and Electron's native dependency rebuild passed on macOS arm64
- `npm test -- --runInBand`: 8 suites and 83 tests passed; preload lint, type checking, and the signed macOS arm64 zip/DMG build passed
- [Three-platform CI run](https://github.com/HASEL-UZH/PersonalAnalytics/actions/runs/34951230610): installation, tests, and installer builds passed on Windows, macOS arm64, and macOS x64 using `actions/setup-node@v5`. This was run from a test-only branch with publishing disabled so the draft release was not changed.
- Packaged macOS arm64 smoke test with a separate test database: Onboarding, Retrospection with seeded Coding/Email activity and a saved self-report, About, Active Times, and Self-Reflection settings rendered; submitting a self-reflection response persisted successfully

Windows and Intel packaged UIs were not exercised interactively. `npm audit --omit=dev` reports 12 production advisories, including one critical advisory in the unchanged `tar` dependency chain; those existing dependency issues are outside #520.

Closes #520


src/electron/electron/preload/index.ts
```diff
@@ -1,25 +1,7 @@
...
-      };
-    } else {
-      obj[key] = value;
-    }
-  }
-  return obj;
-}
+// Electron 29 no longer allows the native ipcRenderer object to cross the context bridge. Expose
+// only the renderer operation used by the application through a plain wrapper instead.
+contextBridge.exposeInMainWorld('ipcRenderer', {
```
grigor-dochev, line 5: Electron 29 no longer allows the complete native `ipcRenderer` object to be passed through `contextBridge`. With the previous implementation, the renderer received an empty object, causing IPC requests to fail and views such as Retrospection to show no data.

The preload now exposes only a plain `invoke` wrapper. I confirmed that `invoke` is the only `ipcRenderer` operation currently used by the renderer, so this preserves existing functionality while following Electron’s recommended security model.

PR #575, opened 2026-07-31
[#571] Fix Retrospection When Switching Timezones

### LINKED ISSUE #571: Bug: Retrospection visualized oddly when switching timezones
When switching the timezone (e.g. from Zurich to Pacific timezone), the retrospection already shows the upcoming day in the evening (e.g. 27.7. evening in Pacific timezone, shows 28.7. which is the case in Zurich, but no interaction data yet).

Can you please check if (a) we're storing UTC values only and (b) if we're then visualizing data in the current local timezone? Is a challenge maybe the 4am cut-off?

Reported on 0.7.0 (AIRbar)

Thank you

grigor-dochev:
# Retrospection: Handle Timezone Changes Correctly (Closes #571)

## Description

Fixes Retrospection showing the wrong calendar day or retaining stale timeline data after the system timezone changes.

Tracker timestamps were already stored in UTC. The problem was that the selected workday was resolved using the Electron main process timezone, which could remain stale after an operating-system timezone change. SQLite's `localtime` conversion then applied that stale timezone when querying the data.

The selected calendar day and its 04:00-to-04:00 workday are now resolved in the renderer's current local timezone. The resulting UTC boundaries are passed through IPC and used consistently for all Retrospection queries and timeline calculations.

### Changes Made

- Stores the selected Retrospection day as a local calendar date instead of a timezone-dependent `Date`
- Resolves the 04:00-to-04:00 workday in the renderer's current timezone
- Passes the resolved UTC start and end boundaries through the Retrospection IPC commands
- Queries UTC timestamps directly instead of relying on SQLite's process-local `localtime` conversion
- Interprets raw Window Activity and User Input timestamps as UTC
- Uses the same resolved workday range for timeline bounds and all dashboard sections
- Detects system timezone changes and reloads Retrospection data
- Moves the selected date to the new local day when Retrospection was displaying Today
- Keeps a manually selected historical calendar date selected while reloading it in the new timezone
- Updates the existing service tests for the new workday-range contract

Closes #571

### DIFF
--- a/src/electron/electron/ipc/IpcHandler.ts
+++ b/src/electron/electron/ipc/IpcHandler.ts
@@ -42,7 +42,10 @@ import type {
   DailySurveySamplingType,
   ExperienceSamplingAnswerType
 } from '../../shared/StudyConfiguration';
-import type { RetrospectionWorkdayRange } from '../../shared/retrospection/Workday';
+import {
+  normalizeRetrospectionWorkdayRange,
+  type RetrospectionWorkdayRange
+} from '../../shared/retrospection/Workday';
 import { DailySurveyTracker } from '../main/services/trackers/DailySurveyTracker';
 import { UsageDataService } from '../main/services/UsageDataService';
 import { UsageDataEventType } from '../enums/UsageDataEventType.enum';
@@ -347,16 +350,16 @@ export class IpcHandler {
   }
 
   private async retrospectionGetActivities(workdayRange: RetrospectionWorkdayRange): Promise<ActivitySessions[]> {
-    return await getActivitySessions(workdayRange);
+    return await getActivitySessions(normalizeRetrospectionWorkdayRange(workdayRange));
   }
 
   private async retrospectionGetActiveHours(workdayRange: RetrospectionWorkdayRange) {
-    return await getActiveHoursInsight(workdayRange);
+    return await getActiveHoursInsight(normalizeRetrospectionWorkdayRange(workdayRange));
   }
 
   private async retrospectionLoadLongestTimeActive(workdayRange: RetrospectionWorkdayRange): Promise<TimeActive | undefined> {
     try {
-      return await getLongestTimeActiveInsight(workdayRange);
+      return await getLongestTimeActiveInsight(normalizeRetrospectionWorkdayRange(workdayRange));
     } catch (error) {
       LOG.error('Error loading longest time active', error);
     }
@@ -366,7 +369,7 @@ export class IpcHandler {
     workdayRange: RetrospectionWorkdayRange
   ): Promise<ActivitySessions[] | undefined> {
     try {
-      return (await getAppUsageSessions(workdayRange))
+      return (await getAppUsageSessions(normalizeRetrospectionWorkdayRange(workdayRange)))
         .sort((a, b) => b.totalDurationMs - a.totalDurationMs)
         .slice(0, 3);
     } catch (error) {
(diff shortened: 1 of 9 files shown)


src/electron/src/views/RetrospectionView.vue
```diff
@@ -36,8 +41,13 @@ const selfReports = ref<ExperienceSamplingDto[]>([]);
 const ACTIVITY_BREAKDOWN_COVERAGE = 0.9;
 const ACTIVITY_BREAKDOWN_MAX_ITEMS = 6;
 const EXCLUDED_ACTIVITY_BREAKDOWN_GROUPS = new Set(['Other', 'Unknown']);
+const TIMEZONE_CHECK_INTERVAL_MS = 1_000;
```
casaout: do we really need to poll timezone changes once a second; isn't it enough to only do the check when opening/focusing a window (or even just the retrospectioN) and possible experience samplings?, or are tz-changes also impacting any other areas?

I know there is a use case when the retrospection remains open when the tz changes, but i think that's rare enough to omit (in favour of simpler code and not having to periodically check the TZ)

src/electron/electron/ipc/IpcHandler.ts
```diff
@@ -345,27 +346,27 @@ export class IpcHandler {
...
-    return await getActiveHoursInsight(new Date(date));
+  private async retrospectionGetActiveHours(workdayRange: RetrospectionWorkdayRange) {
+    return await getActiveHoursInsight(workdayRange);
   }
 
-  private async retrospectionLoadLongestTimeActive(date: Date): Promise<TimeActive | undefined> {
+  private async retrospectionLoadLongestTimeActive(workdayRange: RetrospectionWorkdayRange): Promise<TimeActive | undefined> {
     try {
-      return await getLongestTimeActiveInsight(new Date(date));
+      return await getLongestTimeActiveInsight(workdayRange);
```
casaout: here and elsewhere, shoudl we keep doign new Date(wrokdayRange) in case it ever isn't a date (to avoid failures in the future)?

src/electron/electron/main/services/retrospection/RetrospectionWorkdayData.ts
```diff
@@ -48,7 +48,7 @@ export async function getActiveMinutesForWorkday(date: Date | string): Promise<S
       (entry.scrollDelta ?? 0) > 0 ||
       (entry.movedDistance ?? 0) > 0
     ) {
-      const minuteIndex = getWorkdayMinuteIndex(new Date(entry.tsStart), start);
+      const minuteIndex = getWorkdayMinuteIndex(parseSqliteUtcDateTime(entry.tsStart), workdayRange.start);
       if (minuteIndex >= 0 && minuteIndex < 24 * 60) {
```
casaout: Claude made me aware of this point: with the update, we correctly consider the 23h or 25h workday across a DST boundary with `getRerospectionWorkdayRange`; but here we've hardcoded to 24 hours. is tihs worth to fix (e.g. `Math.round((end - start) / 60_000)`?

casaout (changes requested):

src/electron/electron/main/services/retrospection/RetrospectionWorkdayData.ts
```diff
@@ -33,6 +33,9 @@ export interface RetrospectionWorkdayData {
 export async function getActiveMinutesForWorkday(workdayRange: RetrospectionWorkdayRange): Promise<Set<number>> {
   const workdayStart = formatSqliteUtcDateTime(workdayRange.start);
   const workdayEnd = formatSqliteUtcDateTime(workdayRange.end);
+  const workdayMinuteCount = Math.round(
```
casaout, line 39: maybe worth to breifly comment what it does here or in L55? (i.e. toc larify it's 23 or 25h depending on dailight savings time?)

src/electron/shared/retrospection/Workday.ts
```diff
@@ -35,6 +40,25 @@ export function getRetrospectionWorkdayRange(date: Date | string): Retrospection
   return { start, end };
 }
 
+export function normalizeRetrospectionWorkdayRange(
```
casaout, line 44: please document the method, as I don't tihnk the name really explains what it does

casaout (approved):

